PR #100, opened 2020-10-28
Avoid Network Error from some Ethereum endpoints by specifying the network

fordN:
Resolves [Mission Control Indexer #186](https://github.com/graphprotocol/mission-control-indexer/issues/186)

Specify the network in the JsonRPCProvider connection, so it doesn'thave to detect the network automatically which can cause issues on geth nodes.
Check out the related [Ethers issue](https://github.com/ethers-io/ethers.js/issues/866) for more details.


Jannis:
Obsolete now that we use `StaticJsonRpcProvider`.

PR #185, opened 2021-01-05
indexer-agent: Allocation only claimed if it has collected query fees

fordN:
This PR introduces a minimum query fees collected threshold for claimable allocations. The configurable threshold, defaulting to 0, will serve to avoid sending unnecessary claim transactions for allocations that haven't collected any query fees.

### DIFF
--- a/packages/indexer-agent/src/network.ts
+++ b/packages/indexer-agent/src/network.ts
@@ -65,6 +65,7 @@ export class Network {
   paused: Eventual<boolean>
   isOperator: Eventual<boolean>
   restakeRewards: boolean
+  queryFeesCollectedClaimThreshold: BigNumber
 
   private constructor(
     logger: Logger,
@@ -78,6 +79,7 @@ export class Network {
     paused: Eventual<boolean>,
     isOperator: Eventual<boolean>,
     restakeRewards: boolean,
+    queryFeesCollectedClaimThreshold: BigNumber
   ) {
     this.logger = logger
     this.wallet = wallet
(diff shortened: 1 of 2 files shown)


packages/indexer-agent/src/network.ts
```diff
@@ -206,6 +210,7 @@ export class Network {
       paused,
       isOperator,
       restakeRewards,
+      parseGRT(String(queryFeesCollectedClaimThreshold))
```
Jannis, line 213: Should we use a more explicit number to string conversion here? I'm not sure what `String()` will do to the number.

packages/indexer-agent/src/network.ts
```diff
@@ -391,11 +396,12 @@ export class Network {
       const result = await this.subgraph
         .query(
           gql`
-            query allocations($indexer: String!, $disputableEpoch: Int!) {
+            query allocations($indexer: String!, $disputableEpoch: Int!, $minimumQueryFeesCollected: BigInt!) {
               allocations(
                 where: {
                   indexer: $indexer
                   closedAtEpoch_lte: $disputableEpoch
+                  queryFeesCollected_gt: $minimumQueryFeesCollected
```
Jannis, line 404: I think `gte` may be more intuitive here. If you set the threshold to e.g. 1 GRT but allocations with 1 GRT of fees are not collected, you might end up confused.

Jannis (changes requested):

packages/indexer-agent/src/network.ts
fordN: String() is a essentially a wrapper around `toString` with other fallbacks (for instance if the value is null or not a primitive.) 

As such: either works in this case, but since the threshold value should always be an integer it may be slightly more clear to use the more specific value.toString() conversion.

fordN: I'll go ahead and update to use `.toString()`.

Jannis (approved):

PR #187, opened 2021-01-06
Sync agent inputs independently and reconcile reliably every 2 minutes

Jannis:
Before, we'd only reconcile if any of the inputs changed, but sometimes the agent finishes reconciling but should do another pass sooner than e.g. the next epoch.

This is the case, for instance, if allocations needed to be closed to free up stake for new allocations; in this case the agent will currently reconcile allocations per deployment but if allocations of a later deployment free up stake that could've been used for an earlier deployment, then we need to reconcile again asap.

On the "asap": It should be possible to flag when there is more reconciliation work to do, but identifying whether that is even possible is hard. Simply reconciling again every 2 minutes, without necessarily resyncing all input data is a decently simple workaround.

This commit therefore polls the different pieces of input data independently and adds a 2min-timer into the mix to reconcile more often. This is achieved with a new `join` eventual.

### DIFF
--- a/packages/indexer-agent/package.json
+++ b/packages/indexer-agent/package.json
@@ -17,7 +17,7 @@
     "graph-indexer-agent": "bin/graph-indexer-agent"
   },
   "dependencies": {
-    "@graphprotocol/common-ts": "1.0.1",
+    "@graphprotocol/common-ts": "1.1.1-alpha.2",
     "@graphprotocol/indexer-common": "^0.9.0",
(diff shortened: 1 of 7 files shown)


fordN (approved):
Looking good! 

Since people may refer to this PR for explanation of the feature could you update the description to specify that it is a "2 minute timer"?

Jannis:
@fordN Done, the same message is also in the commit.

Issue #186, opened 2021-01-06
POI null after allocation closed despite full sync


hyperbot16:
I have closed an allocation after 10 days and have noticed that my POI was null despite ensuring that everything subgraph was synced to the latest block at the time of the closure. It seems that I have lost all the rewards for this period because of what is probably a bug. 

Transaction Hash:  0x9c7e7429ccba712ae94e34656fa00375bdf32f0d0eeb139ac99304d7c712e033

`{"level":30,"time":1609893313576,"pid":1,"hostname":"indexer-agent-8684f44768-6k94w","name":"IndexerAgent","component":"Network","indexer":"0xE0C1aF3235Ecee2c085eaD8e918cd394986d2303","operator":"0xA87c60067A47Cb1Ce79FecB2a6FdC2b4F3d01435","allocation":"0x9354Fd32Fe90D33936311D1D7BbFeCe76352497F","deployment":{"bytes32":"0x31edcacc9a53bc8ab4be2eeb0d873409da4c4228cb2d60e4243bd3b4e8af7500","ipfsHash":"QmRhYzT8HEZ9LziQhP6JfNfd4co9A7muUYQhPMJsMUojSF"},"createdAtEpoch":13,"poi":"0x0000000000000000000000000000000000000000000000000000000000000000","createdAtBlockHash":"0x5a340c71b5995c078adf5651a3f4ead963498ac4c0bccff8dda9f069d405c5f8","msg":"Successfully closed allocation"}`

POI at time of allocation creation: 

```
curl -X POST -H "Content-Type: application/json" --data '{"query": "query {proofOfIndexing(subgraph: \"QmRhYzT8HEZ9LziQhP6JfNfd4co9A7muUYQhPMJsMUojSF\",blockHash: \"0x5a340c71b5995c078adf5651a3f4ead963498ac4c0bccff8dda9f069d405c5f8\",indexer: \"0xe0c1af3235ecee2c085ead8e918cd394986d2303\")}"}'  http://localhost:8030/graphql
{"data":{"proofOfIndexing":null}}
```

I main changes I have performed during that period are: 
- Updating to the latest version of the indexer-{service,agent}
- Changing the Ethereum archive node from an openethereum to a geth instance.

Jannis:
Thanks so much for the detailed report, @hyperbot16. I've root-caused it and now understand why this happens:

[...]

PR #191, opened 2021-01-07
indexer-agent: Update to new `proofOfIndexing` API (#186)

Jannis:
This adds passing the epoch start block number to the `proofOfIndexing` query, which is being added to graph-node in https://github.com/graphprotocol/graph-node/pull/2100.

This change is necessary so that we can reliably obtain a POI even if the graph-node block store doesn't have the Ethereum block in question. This fixes the main problem in #186. For another improvement (allowing indexers to deal with legitimate 0x0 POIs by closing allocations with manual intervention), we'll create a separate issue.


fordN (approved):
LGTM! 👍

PR #197, opened 2021-01-14
scripts: Add script to check the POI an epoch/indexer

Jannis:
This is a copy (and slightly improved) version of the script I posted on Discord. I figured it would be good to give this a more permanent place.

### DIFF
--- a/scripts/check-poi.sh
+++ b/scripts/check-poi.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+set -e
+
+QUERY_NODE_STATUS_ENDPOINT=$1
+ETHEREUM_NODE=$2
+EPOCH=$3
+INDEXER=$4
+
+if [[ -z "$QUERY_NODE_STATUS_ENDPOINT" ]] || [[ -z "$EPOCH" ]] || [[ -z "$INDEXER" ]] || [[ -z "$ETHEREUM_NODE" ]]; then
+  echo "Usage: $0 <query-node-status-endpoint> <ethereum-node> <epoch-number> <indexer-address>"
+  echo
+  echo "Example: $0 http://localhost:8030/graphql http://localhost:8545/ 29 0x0000000000000000000000000000000000000000"
+  exit 1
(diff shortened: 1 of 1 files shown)


scripts/check-poi.sh
```diff
@@ -0,0 +1,41 @@
...
+  echo "Example: $0 http://localhost:8030/graphql http://localhost:8545/ 29 0x0000000000000000000000000000000000000000"
+  exit 1
+fi
+
+if ! (which http &>/dev/null); then
+  echo "The 'http' command (https://httpie.io/) must be installed for this script to work."
+  exit 1
+fi
+
+CURRENT_EPOCH=$(http -b post https://gateway.network.thegraph.com/network query='{ graphNetworks { currentEpoch } }' | jq '.data.graphNetworks[0].currentEpoch')
```
abarmat, line 22: Maybe we should fetch this from the contracts in case the current epoch is not stored yet on the subgraph

scripts/check-poi.sh
```diff
@@ -0,0 +1,41 @@
...
+
+HASH=$(echo $BLOCK_DATA | jq '.hash')
+NON_HEX_NUMBER=$(echo "$BLOCK_DATA" | jq '.number' | xargs printf '%d')
+VARIABLES="{\"number\": $NON_HEX_NUMBER, \"hash\": $HASH, \"indexer\": \"$INDEXER\"}"
+echo "Proof of Indexing parameters:" $VARIABLES
+
+echo "---"
+
+http -b post $QUERY_NODE_STATUS_ENDPOINT \
+  query='query poi($number: Int!, $hash: String!, $indexer: String!) { proofOfIndexing(subgraph: "QmRhYzT8HEZ9LziQhP6JfNfd4co9A7muUYQhPMJsMUojSF", blockNumber: $number, blockHash: $hash, indexer: $indexer) }' \
```
abarmat, line 40: Is it ok that the subgraph id is fixed instead as a parameter?

PR #198, opened 2021-01-15
Validate cost models in `setCostModel` to avoid broken cost models

### LINKED ISSUE #182: Missing cost model validation in indexer CLI / indexer management
Right now it's too easy to miss syntax errors in cost models. It shouldn't be hard to integrate Agora and parse/validate cost models before they even make it into the indexer database.

Jannis:
This fixes https://github.com/graphprotocol/indexer/issues/182.

I tried baking this right into the sequelize model so that we could be certain a model is valid before it hits the db, but I always got errors that `CostModel` was not set on `undefined`, so something with the `addon` inside the cost model library wasn't right. Either way, this does what we want it to do.

### DIFF
--- a/packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
+++ b/packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
@@ -3,6 +3,7 @@
 
 import { CostModelVariables, GraphQLCostModel, parseGraphQLCostModel } from '../models'
 import { IndexerManagementResolverContext } from '../client'
+import { compileAsync } from '@graphprotocol/cost-model'
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 const getVariable = (vars: CostModelVariables | null, name: string): any | undefined => {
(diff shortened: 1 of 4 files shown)


packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -68,6 +69,15 @@ export default {
   ): Promise<object> => {
     const update = parseGraphQLCostModel(costModel)
 
+    // Validate cost model
+    try {
+      const modelForValidation = update.model || 'default => 1;'
```
That3Percent, line 74: This might break in the future. For example this cost model:

`default when $UNDER_LOAD => 1;`

Right now what will happen is that this statement will be compiled to...
`default => error;`

But it may be desirable to propagate a warning or error or something instead.

What should happen here is to load the missing data from the database during a partial update to verify the change is valid.

This will work today though, so carry on...

That3Percent (approved):

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
Jannis: Yeah, for now, this is just a little trick to only focus on the variables if no new model was provided. Of course, if at any point the model already in the database becomes invalid due to a breaking change in agora, that wouldn't be caught here.

That3Percent: No, I don't think you understand the point I'm trying to make.

What I'm saying is that the cost model and it's variables may not be able to be validated independently. Whether or not a cost model compiles may depend on the variables. So, by using `{}` here when the model has a partial update may cause an error that would not have been encountered had the variables in the database been used instead.

Issue #183, opened 2020-12-30
Unresolved promise rejections caused by ethers.js


Jannis:
For reference: https://github.com/ethers-io/ethers.js/issues/1208

It looks like after a while, ethers.js always runs into this. One way to work around the problem could be to re-initialize a few of the runtime components, like the `StaticJsonRpcProvider`. This may be a pretty involved workaround, so it would be preferable if ethers.js could be made more robust.

Jannis:
I reworked the agent to recreate the Ethereum provider every 10 minutes but after only 2 minutes it starts throwing `unresolved promise rejections` again. So that idea doesn't cut it as a workaround.

PR #199, opened 2021-01-15
*: Update ethers to 5.0.26 (#183)

Jannis:
ethers 5.0.26 includes a fix for https://github.com/ethers-io/ethers.js/issues/1208, which should remove the unresolved promise rejections we've seen in #183.

Before we merge this, we should also bump common-ts to a new release that includes https://github.com/graphprotocol/common-ts/pull/60.


fordN (approved):
👍

PR #210, opened 2021-02-15
Fix TypeError: Cannot read property 'connect' of undefined

diman-io:
https://github.com/graphprotocol/contracts/issues/449


docs/networks.md
```diff
@@ -69,27 +69,28 @@ indexer. The only requirement is a minimum stake of 100k GRT.
 In order to register for the testnet and have testnet GRT distributed to you,
 please fill out the [testnet registration form](https://airtable.com/shrL1trS84Jf0aawP).
 
+### Preparing tool
```
Jannis, line 72: Can we change this to `### Precompile The Contracts`?

Jannis (changes requested):
Looks good apart from the section title. Could we change that?

tilacog:
It seems that this is no longer an issue. Closing.

PR #212, opened 2021-02-26
Indexer agent: Managing failed or stuck transactions

fordN:
This PR adds gas management tools to the indexer agent to allow better control of gas prices and a more robust transaction management process.

Several gas management startup arguments are introduced to define the transaction resubmission behavior.  
  - **gas-increase-timeout (seconds)**: The time limit after which to resubmit the transaction with more gas.
  - **gas-increase-factor**: Percent to bump the gas when resubmitting.
  - **gas-price-max (gwei)**: The maximum gas price for a resubmitted transaction.
  - **transaction-attempts**: The maximum number of transaction retries to attempt.

Transactions are now retried upon failure; depending on the failure mode the transaction configuration will be adjusted before resubmitting.  For example, if a transaction exceeds the `gas-increase-timeout` it will be resubmitted with the gas price increased by the `gas-increase-factor`.

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -54,10 +54,11 @@ export default {
       })
       .option('gas-bump-time-limit', {
         description:
-          'Time limit for transactions after which the transaction will be resubmitted with more gas (ms)',
+          'Time limit for transactions after which the transaction will be resubmitted with more gas (s)',
         type: 'number',
-        default: 240000,
+        default: 240,
         group: 'Ethereum',
+        coerce: arg => arg * 1000,
       })
       .option('gas-bump-percent', {
         description:
@@ -281,6 +282,11 @@ export default {
             return 'Invalid --indexer-geo-coordinates provided. Must be of format e.g.: 31.780715 -41.179504'
           }
         }
+        if (argv['gas-bump-time-limit']) {
+          if (argv['gas-bump-time-limit'] < 30000) {
+            return 'Invalid --gas-bump-time-limit provided. Must be greater then 30 seconds'
+          }
+        }
         return true
       })
       .option('vector-node', {
@@ -324,6 +330,13 @@ export default {
       async: false,
       level: argv.logLevel,
     })
+
+    if (argv.gasBumpTimeLimit < 90000) {
+      logger.warn(
+        'Gas bump time limit is set to less than 90 seconds. Be careful with this setting as it may ' +
+          'lead to higher gas usage',
+      )
+    }
 
     process.on('unhandledRejection', err => {
       logger.warn(`Unhandled promise rejection`, {
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -52,29 +52,29 @@ export default {
         default: 4000,
         group: 'Ethereum',
       })
-      .option('gas-bump-time-limit', {
+      .option('gas-increase-time-limit', {
         description:
-          'Time limit for transactions after which the transaction will be resubmitted with more gas (s)',
+          'Time (in seconds) after which transactions will be resubmitted with a higher gas price',
         type: 'number',
         default: 240,
         group: 'Ethereum',
         coerce: arg => arg * 1000,
       })
-      .option('gas-bump-percent', {
+      .option('gas-increase', {
         description:
-          'The percent to bump the gas price before resubmitting a stuck transaction (%)',
+          'Percentage by which gas prices are increased when resubmitting transactions',
         type: 'number',
         default: 20,
         group: 'Ethereum',
       })
       .option('gas-price-max', {
-        description: 'The maximum gas price to use for a transaction (gwei)',
+        description: 'The maximum gas price (gwei) to use for transactions',
         type: 'number',
         default: 2000000000,
         group: 'Ethereum',
       })
-      .option('max-retries', {
-        description: 'The maximum number of transaction retries',
+      .option('transaction-attempts', {
+        description: 'The maximum number of transaction attempts',
         type: 'number',
         default: 5,
         group: 'Ethereum',
@@ -282,9 +282,9 @@ export default {
             return 'Invalid --indexer-geo-coordinates provided. Must be of format e.g.: 31.780715 -41.179504'
           }
         }
-        if (argv['gas-bump-time-limit']) {
-          if (argv['gas-bump-time-limit'] < 30000) {
-            return 'Invalid --gas-bump-time-limit provided. Must be greater then 30 seconds'
+        if (argv['gas-increase-time-limit']) {
+          if (argv['gas-increase-time-limit'] < 30000) {
+            return 'Invalid --gas-increase-time-limit provided. Must be at least 30 seconds'
           }
         }
         return true
@@ -331,10 +331,9 @@ export default {
       level: argv.logLevel,
     })
 
-    if (argv.gasBumpTimeLimit < 90000) {
+    if (argv.gasIncreaseTimeLimit < 90000) {
       logger.warn(
-        'Gas bump time limit is set to less than 90 seconds. Be careful with this setting as it may ' +
-          'lead to higher gas usage',
+        'Gas increase timeout is set to less than 90 seconds (~ 6 blocks). This may lead to high gas usage.',
       )
     }
 
@@ -520,10 +519,10 @@ export default {
       argv.allocationClaimThreshold,
       argv.poiDisputeMonitoring,
       argv.poiDisputableEpochs,
-      argv.gasBumpTimeLimit,
-      argv.gasBumpPercent,
+      argv.gasIncreaseTimeLimit,
+      argv.gasIncrease,
       argv.gasPriceMax,
-      argv.maxRetries,
+      argv.transactionAttempts,
     )
     logger.info('Successfully connected to network', {
       restakeRewards: argv.restakeRewards,
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -52,7 +52,7 @@ export default {
         default: 4000,
         group: 'Ethereum',
       })
-      .option('gas-increase-time-limit', {
+      .option('gas-increase-timeout', {
         description:
           'Time (in seconds) after which transactions will be resubmitted with a higher gas price',
         type: 'number',
@@ -60,11 +60,11 @@ export default {
         group: 'Ethereum',
         coerce: arg => arg * 1000,
       })
-      .option('gas-increase', {
+      .option('gas-increase-factor', {
         description:
-          'Percentage by which gas prices are increased when resubmitting transactions',
+          'Factor by which gas prices are increased when resubmitting transactions',
         type: 'number',
-        default: 20,
+        default: 1.2,
         group: 'Ethereum',
       })
       .option('gas-price-max', {
@@ -282,10 +282,13 @@ export default {
             return 'Invalid --indexer-geo-coordinates provided. Must be of format e.g.: 31.780715 -41.179504'
           }
         }
-        if (argv['gas-increase-time-limit']) {
-          if (argv['gas-increase-time-limit'] < 30000) {
-            return 'Invalid --gas-increase-time-limit provided. Must be at least 30 seconds'
+        if (argv['gas-increase-timeout']) {
+          if (argv['gas-increase-timeout'] < 30000) {
+            return 'Invalid --gas-increase-timeout provided. Must be at least 30 seconds'
           }
+        }
+        if (argv['gas-increase-factor'] <= 1.0) {
+          return 'Invalid --gas-increase-factor provided. Must be > 1.0'
         }
         return true
       })
@@ -331,9 +334,17 @@ export default {
       level: argv.logLevel,
     })
 
-    if (argv.gasIncreaseTimeLimit < 90000) {
+    if (argv.gasIncreaseTimeout < 90000) {
       logger.warn(
-        'Gas increase timeout is set to less than 90 seconds (~ 6 blocks). This may lead to high gas usage.',
+        'Gas increase timeout is set to less than 90 seconds (~ 6 blocks). This may lead to high gas usage',
+        { gasIncreaseTimeout: argv.gasIncreaseTimeout / 1000.0 },
+      )
+    }
+
+    if (argv.gasIncreaseFactor > 1.5) {
+      logger.warn(
+        `Gas increase factor is set to > 1.5. This may lead to high gas usage`,
+        { gasIncreaseFactor: argv.gasIncreaseFactor },
       )
     }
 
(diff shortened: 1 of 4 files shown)


packages/indexer-agent/src/network.ts
```diff
@@ -112,16 +124,123 @@ export class Network {
...
+        txConfig.nonceOffset +
+        1
+    } else if (
+      error.message.includes(
+        'Transaction nonce is too low. Try incrementing the nonce.',
+      )
+    ) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      txConfig.nonce! += 1
+    } else if (error.message.includes('Reverted')) {
```
abarmat, line 223: This condition would also catch a valid revert and retry it. Let say that for some reason the agent is closing an already closed allocation, it will revert and then try again with higher gas limit and fail again.

Probably we can know if it was an out of gas revert by looking at the gasUsed being at least ~90% close to the gasLimit as we are adding a buffer of 1.5x for gas limit for the gas estimation.

packages/indexer-agent/src/network.ts
```diff
@@ -112,16 +124,123 @@ export class Network {
...
+      txConfig.nonce! += 1
+    } else if (
+      error.message.includes('Try increasing the fee') ||
+      error.message.includes(
+        'Transaction gas price supplied is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.',
+      ) ||
+      error.message?.includes('timeout exceeded')
+    ) {
+      txConfig.gasPrice =
+        txConfig.gasPrice * txConfig.gasBump > this.gasPriceMax
```
abarmat, line 235: Maybe we can test if txConfig.gasPrice === txConfig.gasPriceMax then avoid resending as it create the same "try increasing fee" exception.

abarmat (changes requested):
Ford, I left a few comments. While reading the nonceOffset part I got a bit confused about how it is used.

packages/indexer-agent/src/network.ts
```diff
@@ -112,16 +124,123 @@ export class Network {
...
+
+        const receipt = await this.ethereum.waitForTransaction(
+          tx?.hash,
+          3,
+          this.gasBumpTimeLimit,
+        )
+
+        if (receipt.status == 0) {
+          const tx = await this.ethereum.getTransaction(receipt.transactionHash)
+          const code = await this.ethereum.call(tx)
```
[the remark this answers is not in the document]
fordN: It should work, I believe a full node keeps state for the latest ~50 blocks then prunes it?

packages/indexer-agent/src/network.ts
fordN: Good idea. I had been thinking that the contract calls beforehand were enough of a check for other revert reasons, but I like this additional protection.  I'll see if I can add a check for gas used being near the limit.

packages/indexer-agent/src/commands/start.ts
```diff
@@ -50,6 +50,32 @@ export default {
         default: 4000,
         group: 'Ethereum',
       })
+      .option('gas-bump-time-limit', {
```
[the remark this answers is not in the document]
fordN: RIght, that would likely lead to a wasteful loop.  I could make the time limit have a minimum of 15 seconds..

packages/indexer-agent/src/network.ts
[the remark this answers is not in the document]
abarmat: Yes, probably 128 blocks, but just wanted to mention in case we need to add some validation.

packages/indexer-agent/src/network.ts
abarmat: Yes, the contracts calls we are doing for gas estimation are good because they will hit a high gas limit error if they will fail, however in the blockchain we can run into some time-sensitive conditions, for example:

1) Does gas estimation on an allocate() transaction and it won't revert.
2) Send a allocate() tx, takes some minutes to mine and when it gets mined some condition make it revert, like the indexer unstaked a chunk.
3) It will enter the revert retry logic.

fordN:
@abarmat I've addressed your comments as well as improved the error logging a bit by adding a few new IndexerError variants.

packages/indexer-agent/src/commands/start.ts
```diff
@@ -52,6 +52,33 @@ export default {
...
+        default: 20,
+        group: 'Ethereum',
+      })
+      .option('gas-price-max', {
+        description: 'The maximum gas price to use for a transaction (gwei)',
+        type: 'number',
+        default: 2000000000,
+        group: 'Ethereum',
+      })
+      .option('max-retries', {
```
Jannis, line 76: Is this the number of retries or the number of attempts? (There's a difference of 1 between them)

I'd call this `transaction-retries` or `transaction-attempts`?

packages/indexer-agent/src/commands/start.ts
```diff
@@ -255,6 +282,11 @@ export default {
             return 'Invalid --indexer-geo-coordinates provided. Must be of format e.g.: 31.780715 -41.179504'
           }
         }
+        if (argv['gas-bump-time-limit']) {
+          if (argv['gas-bump-time-limit'] < 30000) {
+            return 'Invalid --gas-bump-time-limit provided. Must be greater then 30 seconds'
```
Jannis, line 287: Greater than or greater or equal than? 😉

packages/indexer-agent/src/commands/start.ts
```diff
@@ -299,6 +331,13 @@ export default {
       level: argv.logLevel,
     })
 
+    if (argv.gasBumpTimeLimit < 90000) {
+      logger.warn(
+        'Gas bump time limit is set to less than 90 seconds. Be careful with this setting as it may ' +
+          'lead to higher gas usage',
+      )
+    }
+
```
Jannis, line 340: Higher than what?

I'd do
```
Gas increase timeout is set to less than 90 seconds (~ 6 blocks). This may lead to high gas usage.
```

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+      nonce: tx.nonce,
+      gasPrice: tx.gasPrice.toNumber(),
+      gasLimit: tx.gasLimit.toNumber(),
+    }
+
+    logger.info(`Sending transaction`, { tx: tx, attempt: txConfig.attempt })
+
+    while (pending) {
+      if (txConfig.attempt >= this.maxRetries) {
+        logger.warn('Transaction attempt threshold surpassed, moving on', {
```
Jannis, line 156: How about `Transaction retry limit reached, giving up`?

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+            gasPrice: Math.ceil(txConfig.gasPrice),
+            gasLimit: Math.ceil(txConfig.gasLimit),
+          }
+          tx = await this.wallet.sendTransaction(transaction)
+        }
+
+        logger.info(`Transaction pending`, { tx: tx })
+
+        const receipt = await this.ethereum.waitForTransaction(
+          tx?.hash,
```
Jannis, line 183: Under what circumstances would `tx` be `null` or `undefined`? Is this a potential error case we should catch?

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+          3,
+          this.gasBumpTimeLimit,
+        )
+
+        if (receipt.status == 0) {
+          if (receipt.cumulativeGasUsed.toNumber() > txConfig.gasLimit * 0.9) {
```
Jannis, line 189: Why not `> txConfig.gasLimit`?

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+            throw indexerError(IndexerErrorCode.IE050)
+          }
+          throw indexerError(IndexerErrorCode.IE051)
+        }
+
+        logger.info(`Transaction successfully included in block`, {
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          tx: tx?.hash,
```
Jannis, line 197: Same here as with the previous question about `tx` being `null`/`undefined`.

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+    return output!
+  }
+
+  async updateTransactionConfig(
+    logger: Logger,
+    txConfig: Record<string, number>,
+    error: Error | IndexerError,
+  ): Promise<Record<string, number>> {
+    if (error instanceof IndexerError) {
+      if (error.code == 'IE050') {
```
Jannis, line 217: `error.code === IndexerErrorStatus.IE050` would be better, same with the check below.

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+        txConfig.gasLimit = txConfig.gasLimit * txConfig.gasBump
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        txConfig.nonce! += 1
```
Jannis, line 220: When could `nonce` be `null` or `undefined`?

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+      } else if (error.code == 'IE051') {
+        throw indexerError(IndexerErrorCode.IE051)
```
Jannis, line 222: Why not just re-throw `error`?

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+        error.message.includes(
+          'Transaction nonce is too low. Try incrementing the nonce.',
+        )
+      ) {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        txConfig.nonce! += 1
+      } else if (
+        error.message.includes('Try increasing the fee') ||
+        error.message.includes(
+          'Transaction gas price supplied is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.',
```
Jannis, line 245: Should we really check for the entire message? I'm tempted to suggest matching like
```js
error.message.match(/gas price is too low/)
```

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+        ) ||
+        error.message?.includes('timeout exceeded')
+      ) {
+        if (txConfig.gasPrice == this.gasPriceMax) {
```
Jannis, line 249: Just to be very sure, I'd do `txConfig.gasPrice >= this.gasPriceMax`.

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+          throw indexerError(IndexerErrorCode.IE052)
+        } else {
+          txConfig.gasPrice =
+            txConfig.gasPrice * txConfig.gasBump > this.gasPriceMax
+              ? this.gasPriceMax
+              : txConfig.gasPrice * txConfig.gasBump
```
Jannis, line 255: A more elegant way would be
```ts
txConfig.gasPrice = Math.min(this.gasPriceMax, txConfig.gasPrice * txConfig.gasBump)
```

packages/indexer-agent/src/network.ts
```diff
@@ -120,16 +133,134 @@ export class Network {
...
+        }
+      }
+    }
+    logger.warning('Error sending transaction, retrying', {
```
Jannis, line 259: We've been fairly consistent beginning all error messages with `Failed to`, which helps searching the logs. Could we change this to `Failed to send transaction, retrying`?

packages/indexer-agent/src/network.ts
```diff
@@ -357,7 +496,6 @@ export class Network {
         )
         .toPromise()
 
-      // this.logger.info('YAYYYYY', { result: result })
```
Jannis, line 360: 🤣

packages/indexer-common/src/errors.ts
```diff
@@ -112,6 +115,9 @@ export const INDEXER_ERROR_MESSAGES: Record<IndexerErrorCode, string> = {
   IE047: 'Failed to mark transfer as failed',
   IE048: 'Failed to withdraw query fees for allocation',
   IE049: 'Failed to clean up transfers for allocation',
+  IE050: 'Transaction reverted: gas limit likely hit (cumGasUsed > gasLimit * 0.9)',
```
Jannis, line 118: `cumGasUsed`? Why `likely`?

packages/indexer-common/src/errors.ts
```diff
@@ -112,6 +115,9 @@ export const INDEXER_ERROR_MESSAGES: Record<IndexerErrorCode, string> = {
...
+  IE051: 'Transaction reverted: reason unknown',
+  IE052: 'Transaction aborted: max gas price hit',
```
Jannis, line 120: Let's spell things out: `Transaction aborted: maximum configured gas price reached`

Jannis (changes requested):

packages/indexer-agent/src/commands/start.ts
fordN: Attempts. Good point,  updating the name

packages/indexer-agent/src/network.ts
fordN: It will likely never catch any cases for that conditional.  In practice the nodes appear to pause execution of a transaction before hitting the limit, likely before the next operation that will bring it over the limit.

packages/indexer-agent/src/network.ts
fordN: The `ContractTransaction` struct used to initialized the txConfig doesn't require nonce, though it should have it at that point.

packages/indexer-common/src/errors.ts
fordN: Cumulative, I'll update that to not use the abbreviation. 

It's possible though unlikely that the transaction failed for another reason.

packages/indexer-agent/src/network.ts
fordN: It really shouldn't ever be `null` or `undefined`. I'll look into whether I still need this.

packages/indexer-agent/src/network.ts
fordN: I've updated to remove this non-null assertion; it was no longer necessary.

packages/indexer-agent/src/network.ts
fordN: I believe that was a vestige of an earlier version that instantiated the tx object as undefined.   Removed.

packages/indexer-agent/src/network.ts
fordN: I believe that was a vestige of an earlier version that instantiated the tx object as undefined.   Removed.

packages/indexer-agent/src/commands/start.ts
```diff
@@ -52,6 +52,33 @@ export default {
...
       })
+      .option('gas-bump-time-limit', {
+        description:
+          'Time limit for transactions after which the transaction will be resubmitted with more gas (s)',
+        type: 'number',
+        default: 240,
+        group: 'Ethereum',
+        coerce: arg => arg * 1000,
+      })
+      .option('gas-bump-percent', {
```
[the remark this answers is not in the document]
Jannis: I thought about this one some more and I think we should make it `gas-increase-factor` and expect values `> 1.0`.

Jannis (approved):
I made a small change renaming two arguments (see the latest commit). Everything looks good now.

PR #214, opened 2021-03-16
Indexer-agent: Monitor submitted proofs of indexing for potential disputes

fordN:
This PR adds a step to the reconciliation loop of the indexer-agent to check on-chain proofs of indexing against their data to identify potential disputes.  

The agent looks for any closed allocations that are within the dispute period and checks the allocations POI against reference POIs generated from their graph node (if available); if the reference POIs do not match the allocation the allocation is flagged as a potential dispute and saved in the indexer database.  In addition, a `disputes get` command has been added to the indexer-cli for fetching the allocations that have been flagged as potential disputes.

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -372,13 +372,6 @@ class Agent {
   async identifyPotentialDisputes(
     disputableAllocations: Allocation[],
   ): Promise<void> {
-    // const uniqueCloseEpochNumbers: number[] = [...new Set(disputableAllocations.map(allocation => allocation.closedAtEpoch))]
-    // const epochStartHashes: Record<number, string> = {}
-    // for (const epochNumber of uniqueCloseEpochNumbers) {
-    //   const currentEpochStartBlock = await this.network.contracts.epochManager.currentEpochBlock()
-    //   epochStartHashes[epochNumber] = (await this.network.ethereum.getBlock(epochNumber)).hash
-    // }
-
     const uniqueRewardsPools: RewardsPool[] = await Promise.all(
       [
         ...new Set(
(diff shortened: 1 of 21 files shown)


packages/indexer-agent/src/agent.ts
```diff
@@ -347,6 +369,105 @@ class Agent {
...
+          const dispute: POIDisputeAttributes = {
+            allocationID: allocation.id,
+            allocationIndexer: allocation.indexer,
+            allocationAmount: allocation.allocatedTokens.toString(),
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            allocationProof: allocation.poi!,
+            closedEpoch: allocation.closedAtEpoch,
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            closedEpochReferenceProof: rewardsPool!.referencePOI!,
+            closedEpochStartBlockHash: allocation.closedAtBlockHash,
```
abarmat, line 440: Shouldn't this be `allocation.closedAtEpochStartBlockHash` ?

packages/indexer-agent/src/network.ts
```diff
@@ -435,28 +446,167 @@ export class Network {
...
+    deployments: SubgraphDeploymentID[],
+    minimumAllocation: number,
+  ): Promise<Allocation[]> {
+    let dataRemaining = true
+    let allocations: Allocation[] = []
+
+    try {
+      const zeroPOI = utils.hexlify(Array(32).fill(0))
+      // TODO: Use disputable epochs period value from the network instead of hardcoding to 2 epochs
+      const disputableEpoch = currentEpoch.toNumber() - 2
```
abarmat, line 470: Interesting. We can set a constant for this, or even better to read from a config variable. Disputable allocations will depend very much on a social contract called the Arbitration Charter, probably we want let indexers to change this number easily.

packages/indexer-agent/src/agent.ts
```diff
@@ -256,6 +259,20 @@ class Agent {
     )
     this.logger.info(`Waiting for network data before reconciling every 120s`)
 
+    const disputableAllocations = join({
+      currentEpoch,
+      activeDeployments,
```
abarmat, line 264: This means we are checking disputable allocations only using the indexer active subgraph deployments, is that right?

fordN: correct, the indexer will only be able to generate reference POIs for subgraphs they are actively indexing.

packages/indexer-agent/src/network.ts
fordN: I've added a startup parameter to configure this constant, `--poi-disputable-epochs`.

abarmat (approved):
🚀

PR #216, opened 2021-03-23
Indexer-agent, indexer-cli: Improve allocation POI monitoring efficiency, allow filtering of disputes from cli

fordN:
This PR updates the allocation POI monitoring algorithm to avoid checking allocation POIs that have already been identified as potential disputes and stored in the indexer db (`POIDisputes` table).  In addition the indexer-cli has been updated, so disputes can be filtered when fetching them, new structure `indexer disputes get <status> <minimumAllocationClosedEpoch>`.

### DIFF
--- a/packages/indexer-agent/src/indexer.ts
+++ b/packages/indexer-agent/src/indexer.ts
@@ -178,7 +178,7 @@ export class Indexer {
           if (result.error) {
             throw result.error
           }
-          this.logger.debug('Reference Proof of indexing generated', {
+          this.logger.debug('Reference POI generated', {
             indexer: this.indexerAddress,
             subgraph: deployment.ipfsHash,
             block: block,
@@ -386,8 +386,8 @@ export class Indexer {
             }
           `,
           {
-            status: status,
-            minClosedEpoch: minClosedEpoch,
+            status,
+            minClosedEpoch,
           },
         )
         .toPromise()
--- a/packages/indexer-agent/src/indexer.ts
+++ b/packages/indexer-agent/src/indexer.ts
@@ -149,13 +149,6 @@ export class Indexer {
     try {
       return await pRetry(
         async attempt => {
-          this.logger.info(`Query proof of indexing`, {
-            indexer: this.indexerAddress,
-            subgraph: deployment.ipfsHash,
-            block: block,
-            attempt,
-          })
-
           const result = await this.statuses
             .query(
               gql`
@@ -185,8 +178,12 @@ export class Indexer {
           if (result.error) {
             throw result.error
           }
-          this.logger.info('Proof of indexing returned', {
+          this.logger.debug('Reference Proof of indexing generated', {
+            indexer: this.indexerAddress,
+            subgraph: deployment.ipfsHash,
+            block: block,
             proof: result.data.proofOfIndexing,
+            attempt,
           })
 
           return result.data.proofOfIndexing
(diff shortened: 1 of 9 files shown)


packages/indexer-agent/src/indexer.ts
```diff
@@ -185,8 +178,12 @@ export class Indexer {
           if (result.error) {
             throw result.error
           }
-          this.logger.info('Proof of indexing returned', {
+          this.logger.debug('Reference Proof of indexing generated', {
```
Jannis, line 181: The capitalization here is weird. `Reference POI generated` or `Reference Proof Of Indexing generated` maybe?

packages/indexer-cli/src/commands/indexer/disputes/get.ts
```diff
@@ -24,8 +23,7 @@ module.exports = {
     const { print, parameters } = toolbox
 
     const { h, help, o, output } = parameters.options
-
-
+    const [status, minAllocationClosedEpoch] = parameters.array || []
```
Jannis, line 26: I see that `--merged` is part of the help text of this subcommand. I think that needs to be removed.

packages/indexer-cli/src/commands/indexer/disputes/get.ts
```diff
@@ -6,8 +6,7 @@ import { createIndexerManagementClient } from '../../../client'
 import {disputes, printDisputes} from '../../../disputes'
 
 const HELP = `
-${chalk.bold('graph indexer disputes get')} [options] all
-${chalk.bold('graph indexer disputes get')} [options] <status>
+${chalk.bold('graph indexer disputes get')} [options] <status> <minimumAllocationClosedEpoch>
```
Jannis, line 9: What are the supported status values? I think it's worth documenting them. And in case they are `"Potential"` etc. we should make them all lowercase because that's more ergonomic.

packages/indexer-common/src/indexer-management/resolvers/poi-disputes.ts
```diff
@@ -15,10 +16,20 @@ export default {
...
 
   disputes: async (
-    _: {},
+    { status, minClosedEpoch }: { status: string; minClosedEpoch: number },
     { models }: IndexerManagementResolverContext,
   ): Promise<object | null> => {
     const disputes = await models.POIDispute.findAll({
+      where: {
+        [Op.and]: [
+          { status: status },
```
Jannis, line 25: Just `status` will do here, `status: status` is not necessary

Jannis (changes requested):

packages/indexer-cli/src/commands/indexer/disputes/get.ts
fordN: `potential`, `valid`, and `pending` but only `potential` is used at the moment.  

I've updated to use lowercase `potential` which I agree is preferable.

Jannis: Should we document the potential status values in the help text?

Jannis (approved):
My only request would be that we document the `<status>` values in the help text.

packages/indexer-cli/src/commands/indexer/disputes/get.ts
```diff
@@ -6,13 +6,13 @@ import { createIndexerManagementClient } from '../../../client'
...
 
+  <status>  potential|pending|valid
```
Jannis, line 11: Can we add an empty line above? to separate this from the command line summary?

Jannis (changes requested):

packages/indexer-cli/src/commands/indexer/disputes/get.ts
Jannis: Actually, there is one already, nevermind.

Jannis (approved):
Good to go!

PR #218, opened 2021-03-29
New payments solution

Jannis:
This introduces a new payments solution based on the /vector/ state channels network implementation.

There are a bunch of benefits from this new approach:

- A simple, layered design (receipts -> transfers -> state channels -> contracts).
- Better performance / throughput.
- No risk of out-of-sync channels between gateways and indexers.

Overall, this solution is significantly more efficient and robust than what was implemented before.


That3Percent (approved):
Reviewed in Edge & Node.

PR #220, opened 2021-03-30
Added indexer-native

That3Percent:
This adds async native signature verification methods to speed up `indexer-service`


packages/indexer-native/native/src/attestation.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl StructType for Receipt {
+    const TYPE_NAME: &'static str = "Receipt";
+    fn visit_members<T: MemberVisitor>(&self, visitor: &mut T) {
+        visitor.visit("requestCID", &self.request_cid);
+        visitor.visit("responseCID", &self.response_cid);
+        visitor.visit("subgraphDeploymentID", &self.subgraph_deployment_id);
+    }
+}
```
Jannis, line 83: I'm not entirely sure I spotted a use of this, but it may not be obvious?

packages/indexer-native/native/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+use secp256k1::{recovery::RecoverableSignature, Message, PublicKey, Secp256k1, VerifyOnly};
+use std::sync::Arc;
+mod attestation;
+mod signature_verification;
+
+use attestation::AttestationSigner;
+use primitive_types::U256;
+use signature_verification::SignatureVerifier;
+use std::convert::TryInto as _;
+type Address = [u8; 20];
```
Jannis, line 21: nit: Could use an empty line above to separate it from the imports.

packages/indexer-native/lib/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+    v: number,
+    r: string,
+    s: string,
+}
+
+/*
+    Signs attestations using the given signing data
+*/
+export class NativeAttestationSigner {
+    constructor(chainId: string | number, disputeManagerAddress: string, privateKey: string, subgraphDeploymentId: string)
```
Jannis, line 32: Why is `chainId` `string | number`? Is the string for a hex representation? I don't think we have those anywhere. It's always e.g. `"rinkeby"` or `4` and I doubt you accept `"rinkeby"` here.

packages/indexer-native/native/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+pub type AttestationSignerProxy = Proxy<AttestationSigner>;
+impl AttestationSignerImpl for AttestationSignerProxy {}
+pub trait AttestationSignerImpl: Sized + From<AttestationSigner> {
+    fn _init<'c>(cx: &mut CallContext<'c, JsUndefined>) -> SafeResult<Self> {
+        let chain_id: U256 = cx.arg(0)?;
+        let mut chain_id_bytes = [0u8; 32];
+        chain_id.to_big_endian(&mut chain_id_bytes);
+        let chain_id = eip_712_derive::U256(chain_id_bytes);
```
Jannis, line 57: I wonder if this can't be simplified to just take a regular `u32` argument and convert it to a U256. It's unlikely that we'll see chain IDs like `10000000000000000000000000000` that require a U256 — other than for the EIP-712 construction of course.

packages/indexer-native/native/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+// So it's just being used as a thin wrapper to the actual code.
+declare_types! {
+    pub class NativeSignatureVerifier for SignatureVerifierProxy {
+        init(mut cx) { SignatureVerifierProxy::_init(&mut cx).finish(cx) }
+        method verify(mut cx) { SignatureVerifierProxy::_verify(&mut cx).finish(cx).map(|v| v.upcast()) }
+    }
+
+    pub class NativeAttestationSigner for AttestationSignerProxy {
+        init(mut cx) { AttestationSignerProxy::_init(&mut cx).finish(cx) }
+        method createAttestation(mut cx) { AttestationSignerProxy::_create_attestation(&mut cx).finish(cx).map(|v| v.upcast()) }
```
Jannis, line 92: Out of curiosity: what does `v.upcast()` do?

packages/indexer-native/native/src/signature_verification.rs
```diff
@@ -0,0 +1,55 @@
...
+            .recover(&message, &signature)
+            .map_err(|_| "Failed to recover signature")?;
+
+        match self.signer.load().as_ref() {
+            // If we already have the public key we can do the fast path.
+            Signer::PublicKey(signer) => Ok(signer == &recovered_signer),
+            // If we don't have the public key, but have the address instead
+            // we derive the address from the recovered key. If it's a match
+            // then we can save the public key for the next time avoiding
+            // running keccak on every verification.
```
Jannis, line 35: Nice optimization. Makes you think twice but I get it.

packages/indexer-service/package.json
```diff
@@ -28,7 +28,8 @@
     "@connext/vector-utils": "0.2.1",
     "@google-cloud/profiler": "4.1.1",
     "@graphprotocol/common-ts": "1.3.3",
-    "@graphprotocol/indexer-common": "^0.12.0",
+    "@graphprotocol/indexer-common": "^0.13.0",
+    "@graphprotocol/indexer-native": "file:../indexer-native",
```
Jannis, line 32: I'm going to fix this one before we merge, but let's first address the review.

packages/indexer-service/src/allocations.ts
```diff
@@ -109,18 +110,23 @@ export interface EnsureAttestationSignersOptions {
...
   allocations: Eventual<Allocation[]>
   wallet: Wallet
+  chainId: number
+  disputeManagerAddress: string
 }
 
-export type AttestationSignerCache = LRUCache<string, string>
-export type AttestationSignerMap = Map<string, string>
+export type AttestationSignerCache = LRUCache<string, NativeAttestationSigner>
+export type AttestationSignerMap = Map<string, NativeAttestationSigner>
```
Jannis, line 118: What a neat way to slot these in :)

Jannis (changes requested):
Looking very good. My comments are mostly minor.

packages/indexer-native/lib/index.js
```diff
@@ -0,0 +1,36 @@
+var addon = require('../native');
```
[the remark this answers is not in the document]
That3Percent: Hmm. In other projects `yarn` runs prettier but this does not. I'll run it manually.

packages/indexer-native/lib/index.test.js
```diff
@@ -0,0 +1,71 @@
...
+        // When running the tests locally with some debug information, to ensure
+        // that the fast path was transitioned to it only printed 3/4 messages.
+        // Adding this redundant test printed 4/5 messages. Assuming the problem
+        // is just flushing output.
+        await expect(verifyReceipt(receipt1Tampered)).resolves.toEqual(false)
+    })
+
+    test('Attestation', async () => {
+        const mnemonic =
+            'coyote tattoo slush ball cluster culture bleak news when action cover effort'
```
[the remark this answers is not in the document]
That3Percent: Took this from the attestations test in common-ts

packages/indexer-native/native/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[lib]
+name = "indexer_native"
+crate-type = ["cdylib"]
+
+[build-dependencies]
+neon-build = "0.8.0"
+
+[dependencies]
+neon = "0.8.0"
+native-utils = { git = "ssh://git@github.com/edgeandnode/native-utils.git", commit = "cde9333", branch = "main" }
```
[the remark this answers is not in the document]
That3Percent: Yeah. I've got a backlog of these now (eip-712 in particular)

packages/indexer-native/lib/index.test.js
```diff
@@ -0,0 +1,71 @@
...
+            responseCID: "0x448b8ad3a330cf8f269f487881b59efff721b3dfa8e61f7c8fd2480389459ed3",
+            subgraphDeploymentID: "0x4d31d21d389263c98d1e83a031e8fed17cdcef15bd62ee8153f34188a83c7b1c",
+            v: 27,
+            r: "0x702af3e8dec0aab1b29e5663b7ba6843689a55c2c178a26dcce3bc1eeb3a1de9",
+            s: "0x7b24b529fcf92c9426179146bb7bfed6540043e2c30132e59d994a3cc718f2be",
+        }
+        await expect(signer.createAttestation("request", "response")).resolves.toEqual(expected)
+
+        // Ensure throwing errors works at least in one case when a parameter cannot be deserialized
+        expect(() => new NativeAttestationSigner(chainId, "0xab", privateKey, subgraphDeploymentID)).toThrow()
```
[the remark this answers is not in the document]
That3Percent: That would be cool. It's 10pm and I'm tired though... so not cool enough.

I verified that this is correct by having the attestations test in common-ts operate on the same request/response/key/etc and got the same values for v, r, and s. So... it should be good.

packages/indexer-native/native/src/attestation.rs
That3Percent: It's for EIP-712, which is the format used for attestations.

packages/indexer-native/native/src/lib.rs
That3Percent: Done

packages/indexer-native/lib/index.d.ts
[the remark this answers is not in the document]
That3Percent: 3. Support either when parsing.

I'm choosing 2 for expedience today, but will probably implement 3 the next time I'm in the area.

That3Percent: This is because U256 is supported (decimal)

packages/indexer-native/native/src/lib.rs
That3Percent: EIP-712 specifies that chain_id is U256. This is already conforming to the spec so I'd rather not expend effort to make it not conform to the spec and will prefer to leave it.

packages/indexer-native/native/src/lib.rs
That3Percent: In this case goes from `JsString` to `JsValue`. https://neon-bindings.com/docs/type-checking/

packages/indexer-service/src/allocations.ts
That3Percent: Yeah, I kind of lucked out that you already had all the infrastructure I needed written already.

packages/indexer-service/package.json
That3Percent: Seems there are two issues here. I think you're referring to the local path. Another issue is now that when I run yarn it says it can't find version 0.13.0 for indexer-common.

packages/indexer-native/lib/index.d.ts
Jannis: Oh, right. It's fine, I was just wondering if we could simplify call sites to just pass in a regular integer, because I haven't seen any large chain IDs out there. But this is correct, so we shouldn't change it. :)

That3Percent: You already can pass in a regular integer (eg: 1)

Jannis (approved):

PR #225, opened 2021-04-06
*: Update to latest vector, ethers and common-ts release

Jannis:
This updates all of the mentioned dependencies, as well as the vector node image in the example setup.

We're also building docker images from this PR branch (to be removed before merging) to be able to test the updates.


That3Percent (approved):
I don't know anything about GitHub workflows, but LGTM.

PR #229, opened 2021-04-14
indexer-service: Improve allocation syncing

Jannis:
There's two parts to this:

1. Reduce the allocation syncing interval (and make it configurable), from 10s to 120s.
2. Update common-ts to the latest version that includes a fix for `reduce` eventuals suddenly stopping to work.


That3Percent (approved):
Easy peasy lemon squeezy

PR #232, opened 2021-04-23
Avoid rechecking allocations for potential disputes

fordN:
Previously the POI validator only stored allocations that it identified for potential disputes.  This PR updates the loop to store all allocations that have been checked and on subsequent loops avoids checking allocations that have already been validated.

### DIFF
--- a/packages/indexer-agent/src/network.ts
+++ b/packages/indexer-agent/src/network.ts
@@ -223,7 +223,8 @@ export class Network {
       if (
         error.message.includes(
           'Transaction with the same hash was already imported',
-        )
+        ) ||
+        error.message.includes('nonce has already been used')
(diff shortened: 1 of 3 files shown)


packages/indexer-agent/src/network.ts
```diff
@@ -223,7 +223,8 @@ export class Network {
       if (
         error.message.includes(
           'Transaction with the same hash was already imported',
```
abarmat, line 225: Now that I see this, the error "Transaction with the same hash was already imported" means that the transaction we sent is already on the mempool, so maybe in that case we should skip and just wait for it? if we change the nonce we risk to duplicate the transaction

abarmat (approved):

PR #234, opened 2021-04-23
Add detective-only mode

fordN:
Allow the agent to be run in detective-only mode, so one can monitor the network without interacting with it.

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -269,6 +269,12 @@ export default {
         default: false,
         group: 'Disputes',
       })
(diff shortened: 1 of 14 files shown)


abarmat (approved):

packages/indexer-agent/src/commands/start.ts
```diff
@@ -269,6 +269,12 @@ export default {
         default: false,
         group: 'Disputes',
       })
+      .option('focused-detective', {
```
Jannis, line 272: What does 'focused' mean here? I'm not sure that's an intuitive name.

PR #236, opened 2021-04-27
indexer-agent: Don't remove subgraph deployments with active allocations

fordN:
If a subgraph deployment is no longer being synced by the index nodes the agent may not have an up to date enough POI available to close the allocation.  This PR aims to avoid this situation by never removing a subgraph deployment that has an active allocation.


fordN:
Duplicate of #237

PR #239, opened 2021-05-05
indexer-native: (Performance) Use verify instead of recover for signatures when possible

That3Percent:
The `verify` is significantly faster than `recover`. After we have seen the public key once, prefer using `verify`.


Jannis (approved):
LGTM

PR #240, opened 2021-05-06
indexer-agent: Encode receipts for voucher

That3Percent:


packages/indexer-agent/src/query-fees/allocations.ts
```diff
@@ -132,10 +140,23 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
-      // TODO: Encode the receipts batch
-      const encodedReceipts = '...'
+      // Encode the receipt batch to a buffer
+      // [allocationId, receipts[]] (in bytes)
+      const encodedReceipts = new BytesWriter(20 + (receipts.length * 112));
+      encodedReceipts.writeHex(receipts[0].allocation);
+      for (let receipt of receipts) {
+        // [fee, id, signature]
+        const fee = BigNumber.from(receipt.paymentAmount).toHexString();
+        const feePadding = 33 - (fee.length / 2);
```
Jannis, line 150: What's the padding logic / reason here?

packages/indexer-agent/src/query-fees/allocations.ts
```diff
@@ -132,10 +140,23 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        encodedReceipts.writeZeroes(feePadding);
+        encodedReceipts.writeHex(fee);
+        encodedReceipts.writeHex(receipt.id);
+        encodedReceipts.writeHex(receipt.signature);
+      }
+
+      const clientUrl = 'TODO';
```
Jannis, line 158: I _think_ we don't need am extra URL here after constructing the Axios instance. But I may well remember its API wrong. Either way, can fix that after merging.

Jannis (approved):

packages/indexer-agent/src/query-fees/allocations.ts
That3Percent: `toHexString` uses the minimum number of hex pairs to encode. Eg: `1` is `"0x01"`. The encoding we are using is a fixed size hex encoding of U256 which would then require some additional zeroes at the front (big endian).

The number of bytes required is 32. Because of the 0x it seems the hex string has one more byte than it actually does. So if you do `33 - apparentBytes` you would get the number of zeroes required.

Jannis: Fun. :D

PR #241, opened 2021-05-07
Scalar using allocation-based receipts

Jannis:
This introduces allocation-based receipts for Scalar as an alternative to receipts based on Vector transfers. The reason we're doing this is to allow postponing the Vector integration until Vector has sufficiently stabilized and can scale to the degree that we need it to. Allocation-based receipts require no additional infrastructure and have low complexity compared to Vector.

At a high level, the following changes are introduced here:

- Indexer agent and service both have a `--use-vector` option now that is disabled by default.
- Indexer agent and service now support both, Vector-based receipts and allocation-based receipts (the latter being the new default).
- Instead of resolving transfers and withdrawing query fees from the Vector state channels into the staking contract on chain, the indexer agent now collects receipts for closed allocations via an API in the client (usually: gateway), in exchange for a so-called "(query fee) voucher" and then redeems this voucher on chain via a new `AllocationExchange` contractContract.
- New database models were introduced for allocation-based receipts and vouchers.

One thing I'd like to add either as part of this PR but more likely afterwards is metrics for incoming receipts, receipt collection and voucher redeeming.


That3Percent (approved):

Jannis:
@That3Percent All comments addressed, I'll test from this branch before merging.

That3Percent (approved):

PR #244, opened 2021-05-17
Migration fixes / model improvements

Jannis:
This half way between bug and enhancement, as it contains fixes and improvements. 

Firstly, this PR makes it so that only the indexer agent syncs models to the database. This ensures that models are not modified by both, agent and service, in a conflicting way.

Secondly, it makes the transaction that renames `receipts` to `transfer_receipts` more robust. Due to the way sequelize models are defined, they may _still_ be written to the db by the indexer service (e.g. new models are automatically created as tables), and so we need to check whether the `transfer_receipts` table already exists before trying to rename the old `receipts` table. The old migration code failed for a number of indexers on testnet.

_Note: Ideally, we'd copy receipt data where possible, but in this case, the old receipts data is only compatible with Vector and by the time we'll switch back to Vector, the allocations that these receipts were for will be long gone._

The third change in this PR is renaming all `paymentAmount` fields to `fees` and renaming `queryFees` to `collectedFees` to make things a little more clear/obvious.


packages/indexer-service/src/query-fees/allocations.ts
```diff
@@ -131,23 +131,21 @@ export class AllocationReceiptManager implements ReceiptManager {
...
               transaction,
             })
 
             // Don't save over receipts that are already advanced
             if (!isNew) {
-              const storedPaymentAmount = BigNumber.from(
-                state.getDataValue('paymentAmount'),
-              )
-              if (storedPaymentAmount.gte(receipt.paymentAmount)) {
+              const storedPaymentAmount = BigNumber.from(state.getDataValue('fees'))
```
That3Percent, line 141: Rename to `storedFees`.
Might as well do a search for "payment" across the whole package while you're at it.

That3Percent (changes requested):

packages/indexer-service/src/query-fees/allocations.ts
Jannis: Good catch, will do another pass.

That3Percent (approved):

PR #245, opened 2021-05-22
indexer-common: (Bugfix) AsyncCache could perpetually fail

That3Percent:


Jannis (approved):
The explanation makes sense to me, good to merge!

PR #248, opened 2021-05-26
*: Ensure allocation summaries for pre-existing allocations

Jannis:
To track allocations and associated receipts, the agent creates allocation summaries in the db for every allocation it opens. It then links incoming receipts to these based on the allocation ID. The problem is, these allocation summaries are only created when allocations are created. Most, if not all indexers, however, already have allocations from before 0.15.0 and there's no summaries for these.

This commit ensures allocation summaries are created when flushing receipts, to make sure receipts that were created before 0.15.0 (or manually without the agent running) can be flushed to the database.

### DIFF
--- a/.github/workflows/indexer-agent-image.yml
+++ b/.github/workflows/indexer-agent-image.yml
@@ -4,8 +4,7 @@ on:
   push:
     branches:
     - main
-    - jannis/release-prep
-    - jannis/allocation-receipts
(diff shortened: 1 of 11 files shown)


Jannis:
Tested on mainnet with the help of one of the indexers. I will go ahead, merge and prepare a new release due to the urgency in resolving this issue.

PR #252, opened 2021-06-02
Indexer agent: Detect revert reason for failed transactions 

fordN:
This PR updates the transaction management mechanism to improve handling of failed transactions. The agent now fetches the revert reason for failed transactions and only retries the transaction with transaction configuration updates if it was an `out of gas` issue. 

I've also updated the default value for `transaction-attempts` from `5` to `2` to prioritize gas conservation in the default configuration.

### DIFF
--- a/packages/indexer-agent/src/network.ts
+++ b/packages/indexer-agent/src/network.ts
@@ -138,6 +138,7 @@ export class Network {
     const estimatedGas = await gasEstimation()
     const txPromise = transaction(Math.ceil(estimatedGas.toNumber() * 1.5))
     let tx = await txPromise
+    let txRequest: providers.TransactionRequest | undefined = undefined
 
     let txConfig: Record<string, number> = {
       attempt: 1,
@@ -163,7 +164,7 @@ export class Network {
           logger.info('Resubmitting transaction', {
             txConfig: txConfig,
           })
-          const transaction: providers.TransactionRequest = {
+          txRequest = {
             value: tx.value,
             to: tx.to,
             data: tx.data,
@@ -173,7 +174,7 @@ export class Network {
             gasPrice: Math.ceil(txConfig.gasPrice),
             gasLimit: Math.ceil(txConfig.gasLimit),
           }
-          tx = await this.wallet.sendTransaction(transaction)
+          tx = await this.wallet.sendTransaction(txRequest)
         }
 
         logger.info(`Transaction pending`, { tx: tx })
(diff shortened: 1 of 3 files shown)


packages/indexer-agent/src/network.ts
```diff
@@ -206,6 +214,25 @@ export class Network {
...
+      if (e.body.includes('out of gas')) {
+        revertReason = 'out of gas'
+      } else {
+        throw indexerError(IndexerErrorCode.IE051)
+      }
+    }
+    logger.warn('Transaction reverted:', { reason: revertReason })
+    return revertReason
+  }
+
```
Jannis, line 235: Looking at this, I wonder if this is sufficient or if we need to be smarter about detecting the revert reason. See https://github.com/graphprotocol/graph-node/blob/6d9b1ecf1b7b40e16b07c3d465345fea5272a905/chain/ethereum/src/ethereum_adapter.rs#L468-L569 for comparison.

There's also an `eth-revert-reason` that may be worth checking out, rather than rolling our own. https://github.com/authereum/eth-revert-reason

packages/indexer-common/src/errors.ts
```diff
@@ -127,6 +128,7 @@ export const INDEXER_ERROR_MESSAGES: Record<IndexerErrorCode, string> = {
   IE054: 'Failed to collect receipts in exchange for query fee voucher',
   IE055: 'Failed to redeem query fee voucher',
   IE056: 'Failed to remember allocation for collecting receipts later',
+  IE057: 'Transaction reverted: contract requirement not met',
```
Jannis, line 131: What's a "contract requirement"? I feel this is confusing.

Jannis (changes requested):

packages/indexer-agent/src/network.ts
fordN: I tried `eth-revert-reason` and it was getting errors, so I used it as inspiration to implement in our system.

packages/indexer-common/src/errors.ts
fordN: It results from a revert() or require() in the contract.  I felt "contract requirement" sufficiently generally covered both of those cases.  

It'd be nice to include the actual revert reason string, but for now I'm open to suggestions for a general description.

Jannis: How about `Transaction reverted due to failing assertion in contract`? Essentially, that's what `revert()`, `require()` and `assert()` are: assertions.

packages/indexer-agent/src/network.ts
Jannis: Ah, ok. Let's hope this covers enough cases / Ethereum implementations then. :)

packages/indexer-common/src/errors.ts
fordN: True, sounds good. 👍

packages/indexer-agent/src/network.ts
fordN: I've tested with Alchemy. Maybe I should also test with Infura before merging?

Jannis (approved):
LGTM

PR #253, opened 2021-06-03
indexer-service: Add auth-protected `/network` endpoint

Jannis:
This adds a `/network` endpoint that always points to the latest network subgraph deployment.

Ideally, this wouldn't go through the indexer agent at all but that would require some more work. Since this is only intended to be run in the indexers that back the gateway's `/network` endpoint, I suggest we try this and improve if necessary.


promaty (approved):

Jannis:
@promaty I've added a significant amount of changes in the most recent commit. This moves the “local deployment or fall back to endpoint” network subgraph logic into indexer-common, allowing both the agent and service to use it directly.

With the commit before, the service would have had to go through the agent for every /network query and that would have been bad for performance — the service is supposed to scale horizontally but there’s only one agent.

packages/indexer-common/src/indexing-status.ts
```diff
@@ -0,0 +1,76 @@
...
+        )
+        .toPromise()
+      return (
+        result.data.indexingStatuses
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          .map((status: any) => ({
+            ...status,
+            subgraphDeployment: new SubgraphDeploymentID(status.subgraphDeployment),
+          }))
+          .pop()
```
promaty, line 66: why pop?

promaty (approved):

packages/indexer-agent/src/commands/start.ts
```diff
@@ -588,13 +588,30 @@ export default {
...
+    const networkSubgraphDeployment = argv.networkSubgraphDeployment
+      ? new SubgraphDeploymentID(argv.networkSubgraphDeployment)
+      : undefined
+    if (networkSubgraphDeployment !== undefined) {
+      // Make sure the network subgraph is being indexed
+      await indexer.ensure(
+        `${networkSubgraphDeployment.ipfsHash.slice(
+          0,
+          23,
+        )}/${networkSubgraphDeployment.ipfsHash.slice(23)}`,
```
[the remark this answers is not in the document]
Jannis: Because subgraph names are currently limited to `foo/bar` and I think both prefix and suffix have to be shorter than the full IPFS hash would be. We do the same for the subgraphs managed by the indexer agent. It's a bit hacky but we haven't had issues with it so far.

packages/indexer-common/src/indexing-status.ts
Jannis: Because the `indexingStatuses` query returns an array, in this case with either no elements or a single element (the status for that deployment, if available).

PR #259, opened 2021-06-24
indexer-common, indexer-service: Optimize /network execution

Jannis:
When profiling the indexer service performance and memory usage just processing `/network` requests, I noticed that most time and memory is spent and allocated in urql (and wonka).

Since all the `/network` endpoint does is forward to another endpoint (local or elsewhere), we don't really need to have a GraphQL-aware client into the mix. This commit switches network subgraph execution from urql to plain HTTP / axios.


That3Percent (approved):

Jannis:
Tested both cases of using a local subgraph deployment or a separate endpoint, and both worked just fine.

PR #260, opened 2021-06-24
indexer-agent: implement batch rebate claiming using claimMany

chriswessels:
This implements these changes:
- Adds a `claimMany` function to `network.ts`, wrapping the contract interactions and performing the relevant safety checks
- Instead of iterating through claims and calling each one, call `claimMany` and pass in a batch
- Adds `--rebate-claim-threshold` and `--rebate-claim-batch-threshold` as options

Thanks @fordN!

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -206,10 +206,16 @@ export default {
         default: true,
         group: 'Indexer Infrastructure',
       })
-      .option('allocation-claim-threshold', {
-        description: `Minimum query fees (in GRT) received for an allocation to collect and claim them on-chain`,
-        type: 'number',
-        default: 0,
+      .option('rebate-claim-threshold', {
+        description: `Minimum value of query fees for a single allocation (in GRT) in order for it to be included in a batch rebate claim on-chain`,
+        type: 'string',
+        default: '0',
+        group: 'Indexer Infrastructure',
+      })
+      .option('rebate-claim-batch-threshold', {
+        description: `Minimum total value of query fees in an allocation batch (in GRT) before rebates are claimed on-chain`,
+        type: 'string',
+        default: '0',
         group: 'Indexer Infrastructure',
       })
       .option('inject-dai', {
@@ -411,6 +417,18 @@ export default {
       logger.warn(
         `Gas increase factor is set to > 1.5. This may lead to high gas usage`,
         { gasIncreaseFactor: argv.gasIncreaseFactor },
+      )
+    }
+
+    if (argv.rebateClaimThreshold < argv.voucherRedeemMinSingleValue) {
+      logger.warn(
+        `Rebate single minimum claim value is less than voucher minimum redemption value, but claims depend on redemptions`,
+      )
+    }
+
+    if (argv.rebateClaimThreshold === 0) {
+      logger.warn(
+        `Minimum query fee rebate value is 0 GRT, which may lead to claiming unprofitable rebates`,
       )
     }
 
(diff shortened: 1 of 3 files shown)


packages/indexer-agent/src/network.ts
```diff
@@ -557,15 +564,39 @@ export class Network {
...
       if (result.error) {
         throw result.error
       }
 
-      return result.data.allocations.map(parseGraphQLAllocation)
+      const totalFees: BigNumber = result.data.allocations.reduce((total: BigNumber, rawAlloc: { queryFeesCollected: string }) => {
+        return total.add(BigNumber.from(rawAlloc.queryFeesCollected))
+      }, BigNumber.from(0))
+
+      const parsedAllocs: Allocation[] = result.data.allocations.map(parseGraphQLAllocation)
```
fordN, line 579: Move this below the next block so it only executes if the batch value is above the threshold?

packages/indexer-agent/src/network.ts
```diff
@@ -1065,6 +1096,78 @@ export class Network {
     }
   }
 
+  async claimMany(allocations: Allocation[]): Promise<boolean> {
+    const logger = this.logger.child({
```
fordN, line 1100: I would move the allocations property out of this logger instantiation and instead include it in the `Claim tokens from the rebate pool for many allocations` log message. I think it'll be unnecessarily verbose to include entire allocations list in the subsequent log lines that are for specific allocations.

packages/indexer-agent/src/commands/start.ts
```diff
@@ -194,10 +194,22 @@ export default {
...
+        default: '0',
+        group: 'Indexer Infrastructure',
+      })
+      .option('rebate-claim-min-batch-value', {
+        description: `Minimum total value of query fees in an allocation batch (in GRT) before rebates are claimed on-chain`,
+        type: 'string',
+        default: '0',
+        group: 'Indexer Infrastructure',
+      })
+      .option('voucher-redeem-min-single-value', {
```
fordN, line 209: Do you think we can simplify the configs by combining 'voucher-redeem-min-single-value' and 'rebate-claim-min-single-value'?  That is what we currently do, but I suppose some indexers may want to redeem vouchers even though they won't necessarily claim the rebate.

packages/indexer-agent/src/commands/start.ts
```diff
@@ -194,10 +194,22 @@ export default {
         default: true,
         group: 'Indexer Infrastructure',
       })
-      .option('allocation-claim-threshold', {
-        description: `Minimum query fees (in GRT) received for an allocation to collect and claim them on-chain`,
-        type: 'number',
-        default: 0,
+      .option('rebate-claim-min-single-value', {
```
fordN, line 197: Can we simplify this argument name a bit?  

Maybe `rebate-claim-min` and then its batch counterpart would be `rebate-claim-batch-min`?  Another option I've used here in the past is `threshold` which would make it `rebate-claim-threshold` and `rebate-claim-batch-threshold`.

packages/indexer-agent/src/network.ts
```diff
@@ -1065,6 +1096,78 @@ export class Network {
...
+        gasLimit =>
+          this.contracts.staking.claimMany(allocationIds, this.restakeRewards, {
+            gasLimit,
+          }),
+        logger.child({ action: 'claimMany' }),
+      )
+      if (receipt === 'paused' || receipt === 'unauthorized') {
+        return false
+      }
+      logger.info(`Successfully claimed allocations`, { claimedAllocations: allocationIds })
```
fordN, line 1161: Include the count of allocations claimed?  
```
logger.info(`Successfully claimed ${allocationIds.length} allocations`, { claimedAllocations: allocationIds })
```

fordN (changes requested):
Nice! I know you marked this as draft, but it appears close to being ready to merge. I left a few questions and requests in the review comments.

chriswessels:
lgtm!

PR #261, opened 2021-07-01
offchain subgraphs parameter fix

czarly:
this fixes the agent from crashing when the --offchain-subgraphs command line parameter contains the empty string.

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -268,7 +268,7 @@ export default {
           arg.reduce(
             (acc: string[], value: string) => [...acc, ...value.split(',')],
             [],
-          ),
+          )).map((id: string) => id.trim()).filter((id: string) => id.length > 0),
       })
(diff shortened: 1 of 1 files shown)


Jannis:
@czarly I'm not sure I really consider this a bug and more of a misconfiguration.

trader-payne:
@Jannis this is something that we needed for our [mainnet build](https://github.com/StakeSquid/graphprotocol-mainnet-docker/blob/master/docker-compose.yml#L162) to allow us to use it as an env var from outside the compose yml.
This allows us to put the offchain subgraphs in the start file as an optional variable instead of having people to modify their docker-compose.yml file and hardcode the subgraphs there. Our guides are made so they can be as easy and straightforward as possible and without hindering future updates (ie. if you modify your yml file it will conflict with new pulls on future updates).

Jannis:
Ok, I can see how it's useful. I'm happy with the change.

Jannis (approved):

PR #264, opened 2021-07-04
indexer-agent: Retry low fee query fee vouchers later

Jannis:
Add --voucher-expiration to configure the time (in seconds) after which a voucher with too few query fees is permanently deleted. Until then, keep retrying to redeem it to give the indexer a chance to adjust their `--allocation-claim-threshold`.

As requested by Jim on Discord.


fordN (approved):
Nice 👍

PR #267, opened 2021-07-09
indexer-agent: Fix syntax error in offchain-subgraphs option

Theodus:
Error introduced in 5febd002ec4668d9f1679f6d116f53fa836188a0


Jannis (approved):
Oh, shoot, there's a an extra `)` in the original code. Good catch. Confirms once more that it is worth testing even the smallest changes locally.

PR #269, opened 2021-07-13
indexer-agent: Catch subgraph query errors

Theodus:
Some subgraph queries successfully respond with error codes. However,
the return type of the subgraph query was inconsistent with the
structure of the data being returned in the error case. This would
result in error logs that omit the error message because the parsed
response body would result in a type error instead. This change prevents
the type error from occurring such that the log shows the expected error
message.

### DIFF
--- a/packages/indexer-common/src/network-subgraph.ts
+++ b/packages/indexer-common/src/network-subgraph.ts
@@ -1,7 +1,7 @@
 import axios, { AxiosInstance, AxiosResponse } from 'axios'
 import { Eventual, Logger, SubgraphDeploymentID, timer } from '@graphprotocol/common-ts'
 import { DocumentNode, print } from 'graphql'
-import { OperationResult } from '@urql/core'
+import { OperationResult, CombinedError } from '@urql/core'
 import { IndexingStatusResolver, BlockPointer, IndexingError } from './types'
 
 export interface NetworkSubgraphCreateOptions {
@@ -31,6 +31,11 @@ interface NetworkSubgraphOptions {
     graphNodeQueryEndpoint: string
   }
 }
+
+export type QueryResult<Data> = Pick<
+  OperationResult<Data>,
+  'error' | 'data' | 'extensions'
+>
 
 export class NetworkSubgraph {
   logger: Logger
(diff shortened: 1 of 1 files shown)


packages/indexer-common/src/network-subgraph.ts
```diff
@@ -173,7 +173,11 @@ export class NetworkSubgraph {
   ): Promise<OperationResult<Data>> {
     const client = await this.getClient()
     const response = await client.post('', { query: print(query), variables })
-    return JSON.parse(response.data)
+    const data = JSON.parse(response.data)
+    if (data.errors) {
+      throw data.errors
+    }
+    return data
```
Jannis, line 180: The `OperationResult<Data>` type is defined such that it can include `.errors`. How to handle these (whether to throw or anything else) should be up to the caller.

I think what happens is that we were using urql for these queries before and it returned a result with an `.error` field (note the singular vs `.errors` plural). Now all these call sites have outdated error handling.

I would leave this code as it was but change the call sites, such as https://github.com/graphprotocol/indexer/blob/main/packages/indexer-agent/src/network.ts#L377-L379 to check for `data.errors` instead of `data.error`.

Jannis (changes requested):

packages/indexer-common/src/network-subgraph.ts
Theodus: The `OperationResult` type only has an `error` field, not a `errors` field. So checking for `errors` at the call sites fails to compile. Should there be another interface that correctly describes the structure of these results?

Jannis: Oh,  you're right we are sticking to `OperationResult<Data>` _because_ we used urql before and that's what it returned.

In that case let's leave the call sites untouched and do this here:

```ts
const data = JSON.parse(response.data)
if (data.errors) {
  // Create a proper `OperationResult` with a `CombinedError` like urql does
  return {
    data: data.data,
    error: new CombinedError({ graphQLErrors: data.errors }),
  }
}

return data
```

What do you think?

Jannis: (`CombinedError` can be imported from `@urql/core`, I believe)

Theodus: I thought about doing something like that. The issue is constructing the `OperationContext` seems a bit cumbersome, and I haven't found a way of getting it from somewhere.
```ts
return {
  error: new CombinedError({ graphQLErrors: data.errors }),
  operation: {
    key: 0,
    query: query,
    operationName: 'query',
    context: { requestPolicy: 'network-only', url: '?' },
  },
}
```

Theodus: Should I go with this?

Jannis: Eh, yeah, that's not fun. Can we try to define our own `QueryResult` type as e.g. `Pick<OperationResult<Data>, 'error' | 'data'>`, I think that'll be sufficient to make all current callers happy.

Theodus: Done

Jannis (approved):
LGTM 👍🏻

PR #270, opened 2021-07-13
indexer-cli: Remove styling characters from non-table output

Theodus:
The `indexer status` output would include styling characters in the yaml
and json output formats. This would result in fields such as
`"status": "\u001b[32mup\u001b[39m",` instead of `"status": "up",`. This
PR only applies the styling characters when the table output format is
selected.

### DIFF
--- a/packages/indexer-cli/src/commands/indexer/status.ts
+++ b/packages/indexer-cli/src/commands/indexer/status.ts
@@ -136,15 +136,16 @@ module.exports = {
       const keys = Object.keys(pickFields(result.data.indexerEndpoints, []))
       keys.sort()
 
+      const statusUp = outputFormat == 'table' ? chalk.green('up') : 'up'
+      const statusDown = outputFormat == 'table' ? chalk.red('down') : 'down'
+
       data.endpoints = keys.reduce(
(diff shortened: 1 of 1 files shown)


packages/indexer-cli/src/commands/indexer/status.ts
```diff
@@ -136,15 +136,16 @@ module.exports = {
       const keys = Object.keys(pickFields(result.data.indexerEndpoints, []))
       keys.sort()
 
+      const status_up = outputFormat == 'table' ? chalk.green('up') : 'up'
+      const status_down = outputFormat == 'table' ? chalk.red('down') : 'down'
```
Jannis, line 140: nit: JS variables typically use `camelCase`, so I'd rename these to `statusUp` and `statusDown`.

Jannis (changes requested):
Good catch!

packages/indexer-cli/src/commands/indexer/status.ts
Theodus: Fixed

Jannis (approved):
Looking good!

PR #274, opened 2021-07-14
indexer-agent: Fixup formatting

Theodus:
This is just a result of running yarn and committing the results so that they aren't included in unrelated PRs. There should probably be a CI check for this.


Theodus:
If you think it's a good idea to add a CI check for this, I'd be happy to take care of it.

Jannis (approved):
@Theodus I agree, it would be great if you could add checking the formatting in CI so avoid merging any off formatting or linter issues. 😄

PR #275, opened 2021-07-14
ci: Add job to check formatting

Theodus:


Theodus:
Conveniently, I found a directory that wasn't being auto formatted. So that is acting as a test for the CI job

fordN (approved):
Nice! I was noticing some items needing formatting were making it in lately, great to have this auto check.

Issue #256, opened 2021-06-22
Bug: Indexer Agent calls eth node with Authorisation header even if auth not specified.


dm3ch:
I just have found that indexer-agent sends auth header even if there's no auth data in eth node endpoint url.

I have been running `ghcr.io/graphprotocol/indexer-agent:v0.16.0`
I hadn't got any args when I have been executing indexer agent.
I had got following envs:
```
      INDEXER_AGENT_POSTGRES_DATABASE:            graph-node
      INDEXER_AGENT_POSTGRES_HOST:                postgres-postgresql
      INDEXER_AGENT_INDEX_NODE_IDS:               index_node
      INDEXER_AGENT_GRAPH_NODE_QUERY_ENDPOINT:    http://graphprotocol-node-query:8000/
      INDEXER_AGENT_GRAPH_NODE_ADMIN_ENDPOINT:    http://graphprotocol-node-index:8020/
      INDEXER_AGENT_GRAPH_NODE_STATUS_ENDPOINT:   http://graphprotocol-node-query:8030/graphql
      INDEXER_AGENT_PUBLIC_INDEXER_URL:           http://indexer.domain/
      INDEXER_AGENT_NETWORK_SUBGRAPH_ENDPOINT:    https://gateway.testnet.thegraph.com/network
      INDEXER_AGENT_COLLECT_RECEIPTS_ENDPOINT:    https://gateway.testnet.thegraph.com/collect-receipts
      INDEXER_AGENT_ETHEREUM_NETWORK:             rinkeby
      INDEXER_AGENT_DAI_CONTRACT:                 0x9e7e607afd22906f7da6f1ec8f432d6f244278be
      INDEXER_AGENT_NETWORK_SUBGRAPH_DEPLOYMENT:  QmUBRDENCjFt57vVjFc3D4EQoYqXJffKiyFyuGYB2BN8Wi
      INDEXER_AGENT_ETHEREUM: https://some.host
      INDEXER_AGENT_INDEXER_ADDRESS: <deleted>
      INDEXER_AGENT_MNEMONIC: <deleted>
      INDEXER_AGENT_POSTGRES_PASSWORD: <deleted>
      INDEXER_AGENT_POSTGRES_USERNAME: <deleted>
     
```
     
But in mitmproxy output i had found that it adds auth header anyway.
[...]

PR #276, opened 2021-07-15
Pass undefined as auth data if there's empty user and password

dm3ch:
URL parse function provide empty username and password fields if auth data wasn't supplied in auth URL.
But ether.js considers empty username and password as a valid auth data and forms basic auth header with that.
I have faced that some ethereum node operators (without auth) fails to handle such requests.

ether.js maintainers reported that it's not a bug and they are not going to fix this in their lib. So fix for this case should be implemented on graphprotocol/indexer side - https://github.com/ethers-io/ethers.js/pull/1697#issuecomment-869036939

This PR adds checking if username and password fields both empty, then `undefined` would be supplied to ether.js which would disable adding empty auth header to requests.

/fix https://github.com/graphprotocol/indexer/issues/256

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -487,11 +487,22 @@ export default {
       )
     }
 
+    // Prevent passing empty basicAuth info
+    let username;
+    let password;
+    if (providerUrl.username == "" && providerUrl.password == "") {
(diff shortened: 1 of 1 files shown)


fordN (approved):
Thanks for the contribution @dm3ch!
Looks good 👍

Issue #262, opened 2021-07-01
Indexer-agent erroneous retries


fordN:
There have been multiple reports of the indexer-agent attempting actions that lead to reverts, erroneously retrying transactions that were successful, etc..   

Much of this behavior looks similar on the surface, but there are a host of related issues that could be the root cause. To aid in reproducing and debugging I've created this issue, so we have a single place to discuss these related issues and share data.

cryptovestor21:
I am in the middle of a horrible experience today trying to set three allocations with the agent:

![image](https://user-images.githubusercontent.com/59484786/125633442-044c9ec3-9f30-49f2-ba1c-97507cbdca4f.png)

I set three rules in graphcli and went away to work on something else and came back to the above (https://etherscan.io/address/0x2177318680D9E9265470903B6D0ddfCc5EDB054C)

Agent logs: 
[agent_allocateTxnSpam.log](https://github.com/graphprotocol/indexer/files/6816560/agent_allocateTxnSpam.log)

I'm looking at something in the region of $500 worth of wasted gas and my Indexing operation is now paralyzed by the pending transactions, unless I want to spend more gas to cancel them.

I'm losing confidence in using the agent in its current state, would be glad to take on any advice on how I can mitigate this and get operational again without breaking the bank, happy to help in any way I can to mitigate the issue.

torrential-ops:
Here are my etherscan txns to add in the mix. I checked my docker log files and there isn't history far enough back to cover the failed transactions. In the future when closing or allocating I will grab the docker logs at that time. Would it be possible to link the code that is managing the transactions from within the indexer-agent?

https://etherscan.io/address/0xA8410aea388F866BB4B6af465c3Aa5b0F7995aB1

fordN:
Thanks @torrential-ops! 

Here's the [transaction management code](https://github.com/graphprotocol/indexer/blob/main/packages/indexer-agent/src/network.ts#L118-L286)

fordN:
The recent indexer components release (v0.17.0) contains improvements in handling situations where the gas price max is being hit to avoid the situation reported by @cryptovestor21 above.  

Keep the cases coming! I'd like to identify and fix all classes of unnecessary retries and failed transactions from the agent.

begetan:
I just found it is trying to open a broken allocation Opyn Gamma Mainnet. I opened cli to close another transaction using Metamask and proxy contract. This tx stuck in pending state. Right after that it start to spam txsns open allocation.

[...]

PR #279, opened 2021-07-15
indexer-agent: Improve gas-price-max hit failure mode and default settings

fordN:
### Summary 
This PR makes several updates to the gas-price-max configuration.  
- If gas-price-max threshold is hit the agent will now cancel process before sending any transactions, previously it would only be applied for retry transactions.
- Units of gas-price-max input are now in gwei for improved human readability.
- The default gas-price-max value has been bumped to 50 gwei.

### Motivation
[An incident](https://github.com/graphprotocol/indexer/issues/262#issuecomment-879915945) was observed where gas prices were quickly rising and indexer-agent could have saved "wasted" gas if the gas-price-max threshold triggered a cancel earlier.  In addition the gas-price-max value being too low prevented the intended transaction management retry mechanism of retrying with higher gas price.  

Setting a more reasonable default gas-price-max, improving clarity of the argument to users, and failing earlier when the max is hit should avoid these gas wasting scenarios.


packages/indexer-agent/src/commands/start.ts
```diff
@@ -77,8 +77,9 @@ export default {
       .option('gas-price-max', {
         description: 'The maximum gas price (gwei) to use for transactions',
         type: 'number',
-        default: 2000000000,
+        default: 50,
```
abarmat, line 80: In the PR description it says that the default is 20 gwei

fordN: Oops, ya I bumped it up higher since 20 gwei limit will likely get hit pretty often. 

I've now updated the PR comment to reflect the change.

packages/indexer-agent/src/network.ts
```diff
@@ -135,8 +135,13 @@ export class Network {
     let pending = true
     let output: providers.TransactionReceipt | undefined = undefined
 
-    const estimatedGas = await gasEstimation()
-    const txPromise = transaction(Math.ceil(estimatedGas.toNumber() * 1.5))
+    const estimatedGasPrice = await gasEstimation()
+    const paddedGasPrice = Math.ceil(estimatedGasPrice.toNumber() * 1.5)
```
abarmat, line 139: Is the 1.5 always fixed?

fordN: Yeah. 

I think it would be useful to parameterize this; I was thinking that I will make that update in a subsequent PR to avoid bike shedding on this one and because I don't want to keep adding complexity to the transaction management config that is already new to indexers.

fordN: Could be interesting to explore controlling this parameter using "slow", "normal", "fast", "rapid" categories on startup for the entire agent, or by applying them per deployment rule.

PR #280, opened 2021-07-16
indexer-agent: Paginate query for subgraph deployments

Theodus:
The query for subgraph deployments used to determine which are worth
indexing was subject to timing out due to its complexity. This change
paginates the query rather than taking the first 1000. This has the
added benefit of reducing the possibility of missing deployments.


fordN (approved):
👍

PR #284, opened 2021-07-20
indexer-agent: add existing allocation GRT to freeStake check when reallocating

chriswessels:
When reallocating (`closeAndAllocate`), the GRT that we free up when closing the existing allocation becomes immediately available for the new allocation within the batched call.

This PR updates the agent to add the GRT amount of the existing allocation to the check against the Indexer's available tokens.

Without this fix, an indexer would need to have a buffer of free tokens equal to or in excess of the existing allocation or else the agent won't submit the reallocation transaction.


fordN (approved):
👍

PR #285, opened 2021-07-20
indexer-cli: Show indexer deployment status

### LINKED ISSUE #231: Feature request: `indexer status` command shows information on subgraph deployments currently being indexed
## Background
The `indexer status` command output includes data on the rules the indexer has set for controlling their deployments. However, there is not necessarily a 1 to 1 relationship between rules and deployments so an additional table should be added to show all currently indexing subgraphs. 

## Feature
Include an additional table in the `indexer status` output that will show indexing statuses of all the indexer's subgraph deployments.

**Example:**

Indexing Deployments
deployment | synced | health | fatalError | nonFatalErrors | entityCount | node | network | latestBlockNumber | chainHeadBlockNumber | earliestBlockNumber
--- | --- | --- | --- | --- | --- | --- | --- | -- | -- | --
QmbcSvin3TDkWGpGN1vkNjUCCNbkS1ij5AAkA6Ab6zYaap | true | healthy | [] | - | 4677 | node-0 | rinkeby | 8453481 | 8453481 | 7959999

Theodus:
This change adds an Indexer Deployments section to the `indexer status` command. Example output:
```
Indexer Deployments
┌────────────────────────────────────────────────┬────────┬─────────┬────────────┬─────────┬─────────┬───────────────────┬──────────────────────┬─────────────────────┐
│ deployment                                     │ synced │ health  │ fatalError │ node    │ network │ latestBlockNumber │ chainHeadBlockNumber │ earliestBlockNumber │
├────────────────────────────────────────────────┼────────┼─────────┼────────────┼─────────┼─────────┼───────────────────┼──────────────────────┼─────────────────────┤
│ QmVqMeQUwvQ3XjzCYiMhRvQjRiQLGpVt8C3oHgvDi3agJ2 │ false  │ healthy │ -          │ default │ mainnet │ 12421076          │ 12864751             │ 11101016            │
├────────────────────────────────────────────────┼────────┼─────────┼────────────┼─────────┼─────────┼───────────────────┼──────────────────────┼─────────────────────┤
│ QmVEV7RA2U6BJT9Ssjxcfyrk4YQUnVqSRNX4TvYagjzh9h │ false  │ healthy │ -          │ default │ mainnet │ 12594851          │ 12864751             │ 9371102             │
└────────────────────────────────────────────────┴────────┴─────────┴────────────┴─────────┴─────────┴───────────────────┴──────────────────────┴─────────────────────┘
```

Closes #231

### DIFF
--- a/packages/indexer-common/src/indexer-management/client.ts
+++ b/packages/indexer-common/src/indexer-management/client.ts
@@ -116,7 +116,7 @@ const SCHEMA_SDL = gql`
   }
 
   type IndexingError {
-    handler: String!
+    handler: String
     message: String!
   }
 
--- a/packages/indexer-common/src/indexer-management/client.ts
+++ b/packages/indexer-common/src/indexer-management/client.ts
@@ -19,6 +19,7 @@ import costModelResolvers from './resolvers/cost-models'
 import poiDisputeResolvers from './resolvers/poi-disputes'
 import { BigNumber } from 'ethers'
 import { Op, Sequelize } from 'sequelize'
+import { IndexingStatusResolver } from '../indexing-status'
 
 export interface IndexerManagementFeatures {
   injectDai: boolean
@@ -179,6 +180,7 @@ export interface IndexerManagementClientOptions {
   models: IndexerManagementModels
   address: string
   contracts: NetworkContracts
+  indexingStatusResolver: IndexingStatusResolver
   logger?: Logger
   defaults: IndexerManagementDefaults
   features: IndexerManagementFeatures
@@ -231,7 +233,15 @@ export class IndexerManagementClient extends Client {
 export const createIndexerManagementClient = async (
   options: IndexerManagementClientOptions,
 ): Promise<IndexerManagementClient> => {
-  const { models, address, contracts, logger, defaults, features } = options
+  const {
+    models,
+    address,
+    contracts,
+    indexingStatusResolver,
+    logger,
+    defaults,
+    features,
+  } = options
   const schema = buildSchema(print(SCHEMA_SDL))
   const resolvers = {
     ...indexingRuleResolvers,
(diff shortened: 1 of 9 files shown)


fordN (commented):
Code looks good! I would like to test it on my machine to ensure working as expected.

fordN:
@Theodus I'm running into issues when testing this on my machine. 
```
./bin/graph-indexer indexer status       
Error fetching status information: [GraphQL] Cannot return null for non-nullable field IndexingError.handler.
[GraphQL] Cannot return null for non-nullable field IndexingError.handler.
[GraphQL] Cannot return null for non-nullable field IndexingError.handler.
[GraphQL] Cannot return null for non-nullable field IndexingError.handler.
[GraphQL] Cannot return null for non-nullable field IndexingError.handler.
[GraphQL] Cannot return null for non-nullable field IndexingError.handler.```

File: packages/indexer-common/src/indexer-management/client.ts [hunk 1]
```diff
@@ -113,6 +115,32 @@ const SCHEMA_SDL = gql`
     location: GeoLocation
   }
 
+  type IndexingError {
+    handler: String!
```
fordN, line 119: I believe the errors I was seeing were because this is set to non-nullable.  

Though I would also expect that if there was an `IndexingError` that it would have both of these fields. 🤔

Theodus: I've removed the `!`. This is a mistake on my part, since the schema in graph-node doesn't declare the handler as nullable: https://github.com/graphprotocol/graph-node/blob/83f65bca1ff0127693d1ef5bf624e3c4e6e7283f/server/index-node/src/schema.graphql#L69

fordN: Thanks @Theodus!  I'll test again today.

fordN (approved):
Successfully tested on my machine!  Let's get this merged.

Issue #278, opened 2021-07-15
Failed to collect receipts in exchange for an on-chain query fee voucher


86b:
I've been seeing the following (along with 6 other allocation ID's) every time I start up the `indexer-agent` container where it appears the agent is having issues collecting receipts for certain allocations:

```
{
  "level": 30,
  "time": 1626348829631,
  "pid": 1,
  "hostname": "5f1eb7ae2d18",
  "name": "IndexerAgent",
  "component": "AllocationReceiptCollector",
  "allocation": "0x77a99255419C67884A22f47fa599700E0913b46C",
  "receipts": 616,
  "msg": "Collect receipts for allocation"
}
{
  "level": 50,
  "time": 1626348830322,
  "pid": 1,
  "hostname": "5f1eb7ae2d18",
  "name": "IndexerAgent",
  "component": "AllocationReceiptCollector",
  "allocation": "0x77a99255419C67884A22f47fa599700E0913b46C",
  "err": {
    "type": "IndexerError",
    "message": "Failed to collect receipts in exchange for query fee voucher",
    "stack": "IndexerError: Failed to collect receipts in exchange for query fee voucher\n    at Object.indexerError (/opt/indexer/packages/indexer-common/dist/errors.js:139:12)\n    at AllocationReceiptCollector.<anonymous> (/opt/indexer/packages/indexer-agent/dist/query-fees/allocations.js:205:130)\n    at Generator.throw (<anonymous>)\n    at rejected (/opt/indexer/packages/indexer-agent/dist/query-fees/allocations.js:6:65)\n    at processTicksAndRejections (internal/process/task_queues.js:97:5)",
    "code": "IE054",
    "explanation": "https://github.com/graphprotocol/indexer/blob/master/docs/errors.md#ie054",
    "cause": {
      "message": "Request failed with status code 500",
      "stack": "Error: Request failed with status code 500\n    at createError (/opt/indexer/node_modules/axios/lib/core/createError.js:16:15)\n    at settle (/opt/indexer/node_modules/axios/lib/core/settle.js:17:12)\n    at IncomingMessage.handleStreamEnd (/opt/indexer/node_modules/axios/lib/adapters/http.js:260:11)\n    at IncomingMessage.emit (events.js:333:22)\n    at IncomingMessage.EventEmitter.emit (domain.js:482:12)\n    at endReadableNT (_stream_readable.js:1204:12)\n    at processTicksAndRejections (internal/process/task_queues.js:84:21)",
[...]

PR #293, opened 2021-07-30
indexer-agent: Log collectReceipts payload

That3Percent:
For debugging #278


fordN (commented):
Looks good; is this still needed?

That3Percent:
@fordN I guess it's not needed anymore. I find it unlikely we'll have to debug similar issues again.

PR #295, opened 2021-08-06
docs: use ethereum network name instead of id

chriswessels:
A few users have tried specifying the network ID instead of the network name, and reported the following error from `indexer-service`/`indexer-agent`:

```
Error: invalid network (argument="network", value="4", code=INVALID_ARGUMENT, version=providers/5.1.0)
```


fordN (approved):
👍

PR #323, opened 2021-10-29
Add support for unattestable query responses

Theodus:
This change sets the Graph-Attestable header on query results based on
that of the graph-node response. By default this header will be set to
false and no attestation will be provided. See [GIP 20](https://hackmd.io/@CGmldHLrTRaYRnAva-HoPQ/S1W2vMjEt)

### DIFF
--- a/packages/indexer-service/src/server/index.ts
+++ b/packages/indexer-service/src/server/index.ts
@@ -288,6 +288,7 @@ export const createApp = async ({
             res
               .status(response.status || 200)
               .contentType('application/json')
+              .header('Graph-Attestable', String(response.result.attestation != null))
               .send(response.result)
           } catch (error) {
             const err = indexerError(IndexerErrorCode.IE032, error)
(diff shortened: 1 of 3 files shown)


packages/indexer-service/src/server/index.ts
```diff
@@ -288,6 +288,7 @@ export const createApp = async ({
             res
               .status(response.status || 200)
               .contentType('application/json')
+              .header('Graph-Attestable', String(response.result.attestation != null))
```
That3Percent, line 291: We were not intending to surface this header as a part of the public API. It should remain an implementation detail between `graph-node` and `indexer-service` until we can merge the two and then it would go away entirely. The Consumer can check if the attestation != null to see if the response is attestable so it would also be redundant.

That3Percent (changes requested):

That3Percent (approved):

PR #325, opened 2021-11-18
indexer-service: docker entrypoint instead of cmd

chriswessels:
Like indexer-agent's Dockerfile, we should specify an `ENTRYPOINT` in indexer-service's Dockerfile instead of a `CMD`. This is because the `CMD` is overwritten when you specify command line arguments in the `docker run indexer-service ...` form. Currently the image falls back to the ENTRYPOINT defined in the base image, which is `node`, which means that arguments get passed to node instead of the `indexer-service` process.

This PR fixes that, making the default `docker run` behaviour intuitive/run as expected.


fordN (approved):
LGTM!

PR #329, opened 2021-12-01
Indexing rules by subgraph id

fordN:
This PR adds support for defining indexing rules by subgraph id (rather than deployment id). Defining rules at a subgraph level allows the indexer-agent to automatically manage versioning; keeping both the current and previous version synced and allocated towards. 

Changes:
- Update indexing-rule model: deployment --> identifier & identifierType
- CLI interface unchanged but now supports providing subgraph ids where deployment ids were the only supported ids previously.
- Rules by subgraph id are processed into deployment based rules, so the reconciliation logic doesn't need to change.


packages/indexer-agent/src/agent.ts
```diff
@@ -137,11 +182,34 @@ class Agent {
...
+          )
+          .map(rule => rule.identifier!)
+        const subgraphsMatchingRules = await this.network.subgraphs(
+          subgraphRuleIds,
+        )
+        if (subgraphsMatchingRules.length >= 1) {
+          const epochLength =
+            await this.network.contracts.epochManager.epochLength()
+          const blockPeriod = 15
+          const bufferPeriod = epochLength.toNumber() * blockPeriod * 100 // 100 epochs
```
fordN, line 199: Probably going to want to parameterize this, so indexers can set the period for which they continue to support the previous version.

hopeyen (commented):
Looks great 👍  Haven't tried it locally as the indexer-agent still crashes for me, so the comments are only from reading 🥸

packages/indexer-common/src/indexer-management/models/indexing-rule.ts
```diff
@@ -13,7 +13,8 @@ export const INDEXING_RULE_GLOBAL = 'global'
 
 export interface IndexingRuleAttributes {
   id: number
-  deployment: string
+  identifier: string | null
```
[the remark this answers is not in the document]
fordN: Good point! I also can't think of a case where it would need to be `null`.
When I was making this update I originally was using separate nullable `deployment` and `subgraph` properties and I believe this is a vestige of that strategy. I'll update to be just `string`.

packages/indexer-agent/src/agent.ts
```diff
@@ -50,6 +53,48 @@ const uniqueDeployments = (
...
+      if (latestDeploymentVersion) {
+        rule.identifier = latestDeploymentVersion!.deployment.toString()
+        rule.identifierType = SubgraphIdentifierType.DEPLOYMENT
+        const currentTimestamp = Math.floor(Date.now() / 1000)
+        if (
+          latestDeploymentVersion.createdAt >
+          currentTimestamp - previousVersionBuffer
+        ) {
+          const previousDeploymentVersion = ruleSubgraph.versions.find(
+            version => version.version == latestVersion - 2,
```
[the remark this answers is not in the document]
fordN: That could be possible, but this feature would only sync the latest two versions.  If there really are three useful subgraph versions at any given time I feel that is an edge case that should probably be handled with a deployment based rule on the third version (n-2).

I'll update the `toAdd` variable, always nice to use self describing variable names.

hopeyen:
Env all set up :D I'm crashing on the error IE017 below. Does `setIndexingRule` need to be imported in `packages/indexer-agent/src/indexer.ts`?
```"type": "IndexerError",
      "message": "Failed to ensure default global indexing rule",
       "code": "IE017",
      "explanation": "https://github.com/graphprotocol/indexer/blob/master/docs/errors.md#ie017",
      "cause": {
        "type": "b",
        "message": "[GraphQL] column \"identifier\" of relation \"IndexingRules\" does not exist",
        "name": "CombinedError",
        "graphQLErrors": [
          {
            "message": "column \"identifier\" of relation \"IndexingRules\" does not exist",
            "locations": [],
            "path": [
              "setIndexingRule"
            ],
            "extensions": {}
          }
        ]
      }
```

hopeyen:
I think this is probably because my schema isn't updating, so I'm not sure if we would need to revisit `printIndexingRules` or `displayIndexingRules`. My indexer CLI's `rules get all` is missing the identifier column like this
```➜  indexer git:(ford/local-rules-by-gns) graph indexer rules get all                 
┌──────────────────┬─────────────────────┬─────────────────────────┬───────────┬───────────┬──────────┬─────────────────────┬────────┬───────────────┐
│ allocationAmount │ parallelAllocations │ maxAllocationPercentage │ minSignal │ maxSignal │ minStake │ minAverageQueryFees │ custom │ decisionBasis │
├──────────────────┼─────────────────────┼─────────────────────────┼───────────┼───────────┼──────────┼─────────────────────┼────────┼───────────────┤
│ 50,000.0         │ 1                   │ null                    │ null      │ null      │ null     │ null                │ null   │ never         │
├──────────────────┼─────────────────────┼─────────────────────────┼───────────┼───────────┼──────────┼─────────────────────┼────────┼───────────────┤
│ 50,000.0         │ null                │ null                    │ null      │ null      │ null     │ null                │ null   │ never         │
├──────────────────┼─────────────────────┼─────────────────────────┼───────────┼───────────┼──────────┼─────────────────────┼────────┼───────────────┤
│ null             │ null                │ null                    │ null      │ null      │ null     │ null                │ null   │ always        │
├──────────────────┼─────────────────────┼─────────────────────────┼───────────┼───────────┼──────────┼─────────────────────┼────────┼───────────────┤
│ 50,000.0         │ null                │ null                    │ null      │ null      │ null     │ null                │ null   │ always        │
```

hopeyen:
Testing by using deploymentID and subgraphID at the same time. I found that if I `rules set` using one identifier type, and `rules get` using the other one, I got `[GraphQL] Argument "identifier" of required type "String!" was provided the variable "$deployment" which was not provided a runtime value.`

Taking Livepeer for example, with deploymentID `Qmaz1R8vcv9v3gUfksqiS9JUz7K9G8S5By3JYn8kTiiP5K`, subgraphID `0xb31aa1f312595af94f184c5791dc03d6af12fbff-0`. 
```➜  indexer git:(ford/local-rules-by-gns) ✗ ./packages/indexer-cli/bin/graph-indexer indexer rules set 0xb31aa1f312595af94f184c5791dc03d6af12fbff-0 decisionBasis never allocationAmount 12345
┌──────────────────────────────────────────────┬────────────────┬──────────────────┬─────────────────────┬─────────────────────────┬───────────┬───────────┬──────────┬─────────────────────┬────────┬───────────────┐
│ identifier                                   │ identifierType │ allocationAmount │ parallelAllocations │ maxAllocationPercentage │ minSignal │ maxSignal │ minStake │ minAverageQueryFees │ custom │ decisionBasis │
├──────────────────────────────────────────────┼────────────────┼──────────────────┼─────────────────────┼─────────────────────────┼───────────┼───────────┼──────────┼─────────────────────┼────────┼───────────────┤
│ 0xb31aa1f312595af94f184c5791dc03d6af12fbff-0 │ subgraph       │ 12,345.0         │ null                │ null                    │ null      │ null      │ null     │ null                │ null   │ never         │
└──────────────────────────────────────────────┴────────────────┴──────────────────┴─────────────────────┴─────────────────────────┴───────────┴───────────┴──────────┴─────────────────────┴────────┴───────────────┘
➜  indexer git:(ford/local-rules-by-gns) ✗ ./packages/indexer-cli/bin/graph-indexer indexer rules get Qmaz1R8vcv9v3gUfksqiS9JUz7K9G8S5By3JYn8kTiiP5K                                         
[GraphQL] Argument "identifier" of required type "String!" was provided the variable "$deployment" which was not provided a runtime value.```

File: packages/indexer-agent/src/agent.ts [hunk 4]
```diff
@@ -50,6 +53,48 @@ const uniqueDeployments = (
...
+  subgraphs: Subgraph[],
+  previousVersionBuffer: number,
+): IndexingRuleAttributes[] => {
+  const toAdd: IndexingRuleAttributes[] = []
+  rules.map(rule => {
+    const ruleSubgraph = subgraphs.find(
+      subgraph => subgraph.id == rule.identifier,
+    )
+    if (ruleSubgraph) {
+      const latestVersion = ruleSubgraph.versionCount
```
hopeyen, line 67: a bit nit-picking: it might be intuitive if we let `const latestVersion = ruleSubgraph.versionCount - 1` so that below you can find `latestDeploymentVersion` directly with `latestVersion`, and `previousDeploymentVersion` with `latestVersion - 1`?

packages/indexer-agent/src/network.ts
```diff
@@ -468,11 +470,113 @@ export class Network {
...
+                ),
+              } as SubgraphVersion
+            },
+          )
+          return subgraph
+        })
+
+        // In the case of a fresh graph network there will be no published subgraphs, handle gracefully
+        if (results.length == 0) {
+          return []
```
hopeyen, line 549: I remember there was a fix relating to this since _indexer erroneously close allocations_? was it handled somewhere else?

packages/indexer-cli/src/commands/indexer/rules/stop.ts
```diff
@@ -48,24 +47,19 @@ module.exports = {
...
-      print.error(error.toString())
-      process.exitCode = 1
-      return
-    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const [identifier, identifierType] = await processIdentifier(id, { all: false, global: true })
 
     const inputRule = parseIndexingRule({
-      deployment,
+      identifier,
```
hopeyen, line 54: might need to add `identifierType,` here and in some other `parseIndexingRule` such as `start,set,stop,maybe,clear`

hopeyen (changes requested):
I'm trying to understand the handling for `convertSubgraphBasedRulesToDeploymentBased`. If there is an existing deployment based rule and subgraph based rule on the same deployment but with different parameters, should the subgraphBased rule trump the deployment rule?

packages/indexer-agent/src/network.ts
fordN: Good spot, I need to rebase this branch onto latest `main`.

packages/indexer-cli/src/commands/indexer/rules/stop.ts
fordN: The identifier is a unique primary key, so the identifier type shouldn't be needed.  I'll make sure we have some tests for these commands.

hopeyen: I got these when trying the functions without identifier type 
`[GraphQL] Variable "$rule" got invalid value { decisionBasis: "rules", identifier: "0xbbde25a2c85f55b53b7698b9476610c3d1202d88870e66502ab0076b7218f98a" }; Field "identifierType" of required type "IdentifierType!" was not provided.`

fordN: o yep, the model requires them even though they aren't needed in those instances. I've updated these to fix and included tests for all. :)

packages/indexer-agent/src/agent.ts
```diff
@@ -40,6 +43,16 @@ const deploymentInList = (
 ): boolean =>
   list.find(item => item.bytes32 === deployment.bytes32) !== undefined
 
+const deploymentRuleInList = (
```
fordN, line 46: @hopeyen I've created this function to check if a deployment based rule already exists in the db. I use this below in `convertSubgraphBasedRulesToDeploymentBased` to check for existing deployment rules before creating them.  This way any user defined `deployment` rules take precedent over `subgraph` based rules. I'll make note of this order of precedent in the documentation.

PR #333, opened 2021-12-06
Indexer agent logger improvement 

hopeyen:
Adding extra info to the logs
- IPFS hash to `Reconcile deployment allocations`
- Rebate claim amount to `Claim tokens from the rebate pool for allocation`
- Latest block or fatal errors to `Received a null or zero POI for deployment`
- Added attempts to `redeem query fee voucher`

Reduce logs 
- lowered levels for `Allocation rebate rewards`
- added conditions for other logs such as `Reconcile`

Plan for next steps
- more logger stuff
-- ask indexers how they'd change the logs
-- quantify and analyze log messages
- Investigate IE055 in `redeem query fee voucher`

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -50,6 +50,9 @@ const uniqueDeployments = (
   deployments: SubgraphDeploymentID[],
 ): SubgraphDeploymentID[] => deployments.filter(uniqueDeploymentsOnly)
 
+const deploymentIDSet = (deployments: SubgraphDeploymentID[]): Set<string> =>
+  new Set(deployments.map(id => id.display.bytes32))
+
 class Agent {
   logger: Logger
   metrics: Metrics
@@ -335,6 +338,7 @@ class Agent {
         id: allocation.id,
         deployment: allocation.subgraphDeployment.id.display,
         createdAtEpoch: allocation.createdAtEpoch,
+        amount: allocation.queryFeeRebates,
       })),
     })
     if (allocations.length > 0) {
@@ -495,10 +499,15 @@ class Agent {
       }
     }
 
-    this.logger.info('Reconcile deployments', {
-      active: activeDeployments.map(id => id.display),
-      target: targetDeployments.map(id => id.display),
-    })
+    // only show Reconcile when active ids != target ids
+    if (
+      deploymentIDSet(activeDeployments) != deploymentIDSet(targetDeployments)
+    ) {
+      this.logger.info('Reconcile deployments', {
+        active: activeDeployments.map(id => id.display),
+        target: targetDeployments.map(id => id.display),
+      })
+    }
 
     // Identify which subgraphs to deploy and which to remove
     const deploy = targetDeployments.filter(
@@ -598,9 +607,7 @@ class Agent {
           ),
 
           // Whether the deployment is worth allocating towards
-          targetDeployments.find(
-            target => target.bytes32 === deployment.bytes32,
-          ) !== undefined,
+          deploymentInList(targetDeployments, deployment),
 
           // Indexing rule for the deployment (if any)
           rules.find(rule => rule.deployment === deployment.bytes32) ||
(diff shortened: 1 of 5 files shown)


packages/indexer-agent/src/commands/start.ts
```diff
@@ -512,7 +512,7 @@ export default {
     // Prevent passing empty basicAuth info
     let username
     let password
-    if (providerUrl.username == '' && providerUrl.password == '') {
+    if (providerUrl.username == '' || providerUrl.password == '') {
```
fordN, line 515: I think this should stay `&&` since there could be a case where a username string is provided and the password is an empty string.  I believe that is valid.

packages/indexer-agent/src/agent.ts
```diff
@@ -785,11 +796,31 @@ class Agent {
...
     ) {
-      this.logger.error(`Received a null or zero POI for deployment`, {
-        deployment: allocation.subgraphDeployment.id.display,
-        allocation: allocation.id,
-        epochStartBlock,
-      })
+      const indexingStatus = await this.indexer.indexingStatus(allocation.subgraphDeployment.id)
+      const fatalError = indexingStatus.fatalError
+      // can agent check if the error is determinisitic? 
+      // if so inform latest valid POI
```
fordN, line 802: This is a great idea! 
Let's discuss a bit though offline, I believe this is unnecessary with the latest graph node version (just released!) which should always return a POI.

packages/indexer-agent/src/agent.ts
```diff
@@ -609,40 +619,41 @@ class Agent {
...
+          amount: formatGRT(allocation.allocatedTokens),
+        })),
+      },
+    )
 
     // Return early if the deployment is not (or no longer) worth indexing
     if (!worthIndexing) {
       logger.info(
-        `Deployment is not (or no longer) worth indexing, close all active allocations that are at least one epoch old`,
+        `Deployment is not (or no longer) worth indexing, please close allocation if it is from a previous epoch`,
```
fordN, line 655: This log message isn't asking the user to do anything, rather letting the user know it is about to close allocations.  

Maybe something like `Deployment is not (or no longer) worth indexing, closing all active allocations that are at least one epoch old`?

packages/indexer-agent/src/agent.ts
```diff
@@ -609,40 +619,41 @@ class Agent {
...
       rule?.parallelAllocations || 1,
     )
+    const activeAllocationAmount = activeAllocations.reduce(
+      (sum, allocation) => sum.add(allocation.allocatedTokens),
+      BigNumber.from('0'),
+    )
 
-    logger.info(`Reconcile deployment allocations`, {
-      allocationAmount: formatGRT(allocationAmount),
+    logger.trace(
```
fordN, line 634: I like the update to include block current allocation amount and desired allocation amount.  I feel this log is one of the important ones for understanding what the indexer is doing though, think we should keep it `info` or `debug` level?

packages/indexer-agent/src/agent.ts
```diff
@@ -29,6 +29,7 @@ import pMap from 'p-map'
 import pFilter from 'p-filter'
 import { ReceiptCollector } from './query-fees'
 
+// are these lists ordered? why a list?
```
fordN, line 32: Are you proposing a set instead?

packages/indexer-agent/src/agent.ts
```diff
@@ -86,6 +91,7 @@ class Agent {
     this.logger.info(`Connected to Graph node(s)`)
 
     // Ensure there is a 'global' indexing rule
+    // why?
```
fordN, line 94: The global indexing rule is used for defaults on specific indexing rules.

packages/indexer-agent/src/network.ts
```diff
@@ -1432,7 +1431,11 @@ export class Network {
       restakeRewards: this.restakeRewards,
     })
     try {
-      logger.info(`Claim tokens from the rebate pool for allocation`)
+      logger.info(`Claim tokens from the rebate pool for allocation`, {
```
fordN, line 1434: 👍

packages/indexer-agent/src/query-fees/allocations.ts
```diff
@@ -382,6 +382,8 @@ export class AllocationReceiptCollector implements ReceiptCollector {
         },
       )
     } catch (err) {
+      // seems to be trying repeatedly
+      // should this accumulate? or add attemps similar to querying proof of indexing?
```
fordN, line 386: Interesting. 
1. Worth investigating why it's failing? 
2. Could probably improve logging as you suggest, so it's clear to the indexer how many attempts it's tried etc...

fordN (changes requested):
Some nice usability improvements!  
I made a quick pass at reviewing and left some comments and questions.

packages/indexer-agent/src/commands/start.ts
hopeyen: Ah okay! Would it still make logical sense to not check for password at all? ie, what happens when there is empty username + non-empty password?

packages/indexer-agent/src/agent.ts
hopeyen: I was confused by `at least one epoch old`. An allocation that opened last 20 minutes of epoch2 can technically be closed within 30 minutes of it opened, right?

packages/indexer-agent/src/agent.ts
hopeyen: I don't think it matters, just anxious if there was a particular order that I wasn't aware of

packages/indexer-agent/src/commands/start.ts
fordN: See this [PR](https://github.com/graphprotocol/indexer/pull/276/files) where that was added for background. It's important to check for both because that indicates that a user did not enter anything so the headers shouldn't contain username and password.

packages/indexer-agent/src/agent.ts
fordN: O interesting point. One epoch old does imply it has been around for at least 1 epoch. 

How about, `Deployment is not (or no longer) worth indexing, close allocation if eligible`?
or  ``Deployment is not (or no longer) worth indexing, close allocation if opened in a previous epoch`?

packages/indexer-agent/src/agent.ts
hopeyen: maybe a conditional to log only if there was something to reconcile?

packages/indexer-agent/src/query-fees/allocations.ts
hopeyen: Added attempts and gave a limit for 1000 for now... Though I can't seem to get indexer-agent to get to `submitVoucher` after I close allos.

packages/indexer-agent/src/agent.ts
```diff
@@ -582,7 +591,7 @@ class Agent {
 
     this.logger.debug(`Deployments to reconcile allocations for`, {
       number: deployments.length,
-      deployments: deployments.map(deployment => deployment.display),
+      deployments: deploymentIDSet(deployments),
```
fordN, line 594: This will display the bytes32 representation of the deployment ID.  For consistency we may want to display the ipfs hash.

packages/indexer-agent/src/agent.ts
```diff
@@ -795,16 +806,32 @@ class Agent {
     )
 
     // Don't proceed if the POI is 0x0 or null
+    // with the new graph-node, deployment syncing or fatal undeterministic error
```
fordN, line 809: This comment isn't 100% clear to me. 

Maybe something like: "A null PoI means the deployment is either not fully synced or experienced a fatal undeterministic error"

packages/indexer-agent/src/network.ts
```diff
@@ -1673,7 +1677,8 @@ export class Network {
       if (receipt === 'paused' || receipt === 'unauthorized') {
         return false
       }
-      logger.info(`Successfully claimed allocation`)
+      logger.info(`Successfully claimed allocation`, 
+        receipt.logs)
```
fordN, line 1681: Second param should be an object: `logger.info(`Successfully claimed allocation`, { logs: receipt.logs } )`

packages/indexer-agent/src/query-fees/allocations.ts
```diff
@@ -350,41 +351,56 @@ export class AllocationReceiptCollector implements ReceiptCollector {
     const hexPrefix = (bytes: string): string =>
       bytes.startsWith('0x') ? bytes : `0x${bytes}`
 
+    // TODO: investigate the cause of consistent failure
+    const numRetries = 1000
```
fordN, line 355: Have you been seeing consistent failures in this function? 

Can we move this and below to another PR, so this one is focused more on the logs updates?

fordN (approved):
👍

PR #339, opened 2021-12-23
Add option to allocate on network subgraph

Theodus:
Adds an override to allow allocations to the network subgraph. This is useful for the local testing environment.


fordN (approved):
👍

PR #342, opened 2021-12-24
indexer-agent:  Fix subgraph deployments batch query processing

fordN:
Important updates to `network.subgraphDeploymentsWorthIndexing()`:
- Fix bug that led to false empty result returned from `subgraphsWorthIndexing()`.
- If subgraph deployments query returns empty array throw error instead of gracefully returning an empty array.
- Pass error message to IndexerError.cause.

Small improvements to `reconcileDeploymentAllocations()`:
- Update naming  to be more clear: `worthIndexing` --> `worthAllocating`.

### DIFF
--- a/packages/indexer-agent/src/network.ts
+++ b/packages/indexer-agent/src/network.ts
@@ -523,9 +523,11 @@ export class Network {
         const results = result.data.subgraphDeployments
 
         // In the case of a fresh graph network there will be no published subgraphs, handle gracefully
-        if (results.length == 0) {
-          this.logger.warn('No subgraph deployments returned')
-          return []
+        if (results.length == 0 && queryProgress.fetched == 0) {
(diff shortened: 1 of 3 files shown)


packages/indexer-agent/src/network.ts
```diff
@@ -523,9 +523,11 @@ export class Network {
         const results = result.data.subgraphDeployments
 
         // In the case of a fresh graph network there will be no published subgraphs, handle gracefully
-        if (results.length == 0) {
-          this.logger.warn('No subgraph deployments returned')
-          return []
```
fordN, line 528: This was returning an empty `targetDeployments` array if the last batch query happened to be empty. 🤦

hopeyen: ahhh :0!!!!

for fresh graph networks, it should still give the warning below? Could there be one that checks for 'freshness'?

PR #343, opened 2022-01-04
indexer agent & cli - max lifetime options

hopeyen:
### Purpose
Currently, agent uses maximum allocations provided by the network, which is 28 epoches on the mainnet. Some indexers might want the agent to renew the allocations more frequently

### Updates
- Allow indexer to configure deployment-based lifetimes. Use indexer-cli and set with an integer using `allocationLifetime`
- Indexing rules table shows allocationMaxLifetime for each allocation
- The order of importance is id-specific `allocationLifetime`, global `allocationLifetime`, and then network max epoch. Agent allows you to set lifetimes higher than network maximum so be careful with it

### Future considerations
- this can allow indexer to compound the rewards more often and automatic, but `closeAndAllocate` doesn't account for rewards as the transaction are instant


packages/indexer-agent/src/agent.ts
```diff
@@ -616,12 +690,19 @@ class Agent {
...
 
           // Indexing rule for the deployment (if any)
-          rules.find(rule => rule.deployment === deployment.bytes32) ||
-            rules.find(rule => rule.deployment === INDEXING_RULE_GLOBAL),
+          rules
+            .filter(rule => rule.identifier != 'global')
+            .find(
+              rule =>
+                new SubgraphDeploymentID(rule.identifier).bytes32 ===
+                deployment.bytes32,
```
hopeyen, line 698: workaround the identifier types until upstream merges

packages/indexer-agent/src/agent.ts
```diff
@@ -392,6 +396,13 @@ class Agent {
       }) => {
         this.logger.info(`Reconcile with the network`)
 
+        // should indexer have the ability to use a longer lifetime than network max?
+        // or use min like `? Math.min(this.indexer.maxLifetime, maxAllocationEpochs)`
+        // agent can still serve queries without rewards
+        const indexerMaxAllocationEpochs = this.indexer.maxLifetime
+          ? this.indexer.maxLifetime
+          : maxAllocationEpochs
```
hopeyen, line 404: currently, indexer have the ability to use a longer lifetime than network specified, since agent can still serve queries without rewards? or we can take the min `? Math.min(this.indexer.maxLifetime, maxAllocationEpochs)`

packages/indexer-agent/src/agent.ts
```diff
@@ -778,6 +788,10 @@ class Agent {
       ? BigNumber.from(rule.allocationAmount)
       : this.indexer.defaultAllocationAmount
     const desiredNumberOfAllocations = 1
+    // makes sense to me for allocation-specific rule to trumps others
+    const desiredAllocationMaxLifetime = rule?.allocationMaxLifetime
+      ? BigNumber.from(Math.max(1, rule.allocationMaxLifetime - 1))
+      : maxAllocationEpochs
```
hopeyen, line 794: similar to the other comment that it makes sense to me for allocation-specific rule to trumps others

packages/indexer-agent/src/agent.ts
```diff
@@ -877,7 +888,12 @@ class Agent {
       )
     }
 
-    const lifetime = Math.max(1, maxAllocationEpochs - 1)
+    const lifetime = Math.max(
+      1,
+      rule?.allocationMaxLifetime
+        ? rule?.allocationMaxLifetime - 1
+        : maxAllocationEpochs,
```
fordN, line 895: `maxAllocationEpochs - 1`?

fordN (changes requested):

packages/indexer-agent/src/agent.ts
hopeyen: So from the above [calculation for `allocationLifetime`](https://github.com/graphprotocol/indexer/pull/343/files#diff-186fe48e2f955474051fa281f7fd59e1a31551cbe1ea7f3ef4ec6640342c1f92L696) where `lifetime = max epoch - 1`, I was concerned if we want to use `rule.allocationMaxLifetime` for agent to close right before or at the maximum lifetime epoch?

fordN: What do you think about changing the rule property to `allocationLifetime` and if it's not supplied have it default to `maxAllocationEpochs` - 1? Then the behavior would be more explicit and we can just use that value rather than checking if it exists etc...

hopeyen: yes, that makes sense! will update it

packages/indexer-agent/src/agent.ts
```diff
@@ -202,7 +202,11 @@ class Agent {
     )
 
     const maxAllocationEpochs = timer(600_000).tryMap(
-      () => this.network.contracts.staking.maxAllocationEpochs(),
+      () =>
+        // can i pass these 2 numbers into rule creation?
+        this.indexer.maxLifetime
+          ? this.indexer.maxLifetime
+          : this.network.contracts.staking.maxAllocationEpochs(),
```
hopeyen, line 209: @fordN Do you know if I could pass this value to indexing rules for when it initiate a new rule and/or in db migration? I made a placeholder of 27 but ideally it should depend on the network and the specific indexer?

fordN (approved):
👍

PR #348, opened 2022-01-10
indexer-service: Add /subgraphs/health/:deployment route

Jannis:
This route handles only GET requests. It returns an HTTP 200 response if the deployment in question has not failed and is caught up to the chain head. Otherwise, a 500 is returned with different error messages depending on what is wrong.

This can be used to check the health of specific subgraph deployments in environments where the indexing status API is too complicated and simple, URL-based health checks are the only option (e.g. Cloudflare load balancers, k8s readiness probes, status pages).

### DIFF
--- a/packages/indexer-service/src/server/deployment-health.ts
+++ b/packages/indexer-service/src/server/deployment-health.ts
@@ -1,0 +1,89 @@
+import fetch from 'cross-fetch'
+import { Router } from 'express'
+
+export interface DeploymentHealthServerOptions {
+  graphNodeStatusEndpoint: string
+}
+
+export const createDeploymentHealthServer = ({
+  graphNodeStatusEndpoint,
+}: DeploymentHealthServerOptions): Router => {
+  const router = Router()
+
+  // This route returns an HTTP 200 response if the deployment in question
+  // has not failed and is caught up with the chain head; otherwise an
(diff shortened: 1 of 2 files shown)


packages/indexer-service/src/server/deployment-health.ts
```diff
@@ -0,0 +1,89 @@
...
+      return res.status(500).send('Subgraph deployment has failed')
+    }
+
+    const latestBlock = status.chains[0]?.latestBlock
+    const headBlock = status.chains[0]?.chainHeadBlock
+
+    // Check whether the subgraph is caught up with the chain head
+    if (
+      latestBlock?.number === headBlock?.number &&
+      latestBlock?.hash === headBlock?.hash
```
promaty, line 80: I think this should be `headBlock.number - latestBlock?.number < 5` or something because it doesn't execute instantly

hopeyen:
Code lgtm 👍 agree with having small delays in block num.

hopeyen (commented):
agree with promaty that a slight lag should still be healthy, otherwise looks pretty good!

Jannis:
@promaty @hopeyen Comment addressed, this should be ready now.

promaty (approved):

PR #352, opened 2022-01-11
Indexer-agent, indexer-cli: indexing rules applied instantly

hopeyen:
Previously, indexers find agents needing a few minutes before picking up rules, so a common practice was to restart indexer-agent after setting the rules. 

This PR speeds up the frequency of checking indexing rules, and triggers `Reconcile Deployments` within 20 seconds of changing the rules. -> indexers would not need to restart the agent as often. 

Rebased to `ford/local-rule-by-gns`

### DIFF
--- a/packages/indexer-agent/src/network.ts
+++ b/packages/indexer-agent/src/network.ts
@@ -627,9 +627,6 @@ export class Network {
                     id
                   }
                 }
-                subgraph {
-                  id
-                }
               }
             }
           `,
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -209,7 +209,7 @@ class Agent {
       },
     )
 
-    const indexingRules = timer(60_000).tryMap(
+    const indexingRules = timer(20_000).tryMap(
       async () => {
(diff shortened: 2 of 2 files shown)


packages/indexer-agent/src/network.ts
```diff
@@ -537,9 +641,13 @@ export class Network {
           ...results
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
             .filter((deployment: any) => {
+              rules = rules.filter(rule => rule.identifier !== 'global')
               const deploymentRule =
-                rules.find(rule => rule.deployment === deployment.id) ||
-                globalRule
+                rules.find(
+                  rule =>
+                    new SubgraphDeploymentID(rule.identifier).bytes32 ===
```
hopeyen, line 648: temporarily workaround along with `644`

fordN:
Nice, this looks good! Once we get the other PRs merged we can clean this up before merging.

fordN (commented):
Nice, this looks good! Once we get the other PRs merged we can clean this up before merging.

fordN (approved):
👍

PR #354, opened 2022-01-19
indexer-agent, indexer-common: Add live metric for operator ETH balance

Jannis:
This is refreshed every 2 minutes and can be used by indexers to set up alerting for when their operator is running out of ETH to make transactions.

### DIFF
--- a/packages/indexer-agent/src/utils/balance-monitor.ts
+++ b/packages/indexer-agent/src/utils/balance-monitor.ts
@@ -16,7 +16,7 @@ export async function monitorEthBalance(
   wallet: Wallet,
   metrics: Metrics,
 ): Promise<void> {
-  const logger = logger.child({ component: 'ETHBalanceMonitor' })
+  logger = logger.child({ component: 'ETHBalanceMonitor' })
 
   logger.info('Monitor operator ETH balance (refreshes every 120s)')
 
--- a/packages/indexer-agent/src/utils/balance-monitor.ts
+++ b/packages/indexer-agent/src/utils/balance-monitor.ts
@@ -16,6 +16,8 @@ export async function monitorEthBalance(
   wallet: Wallet,
   metrics: Metrics,
 ): Promise<void> {
+  const logger = logger.child({ component: 'ETHBalanceMonitor' })
+
   logger.info('Monitor operator ETH balance (refreshes every 120s)')
 
   const balanceMetrics = registerMetrics(metrics)
@@ -25,6 +27,9 @@ export async function monitorEthBalance(
       const balance = await wallet.getBalance()
       const eth = parseFloat(utils.formatEther(balance))
       balanceMetrics.operatorEthBalance.set(eth)
+      logger.info('Current operator ETH balance', {
+        balance: eth,
+      })
     } catch (error) {
       logger.warn(`Failed to check latest ETH balance`, {
         err: indexerError(IndexerErrorCode.IE059),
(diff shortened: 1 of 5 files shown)


packages/indexer-agent/src/utils/balance-monitor.ts
```diff
@@ -0,0 +1,39 @@
+import { Gauge } from 'prom-client'
+import { Wallet, utils } from 'ethers'
+
+import { Logger, Metrics, timer } from '@graphprotocol/common-ts'
+import { indexerError, IndexerErrorCode } from '@graphprotocol/indexer-common'
+
+interface EthBalanceMetrics {
+  operatorEthBalance: Gauge<string>
+}
```
promaty, line 9: You can avoid writing this interface by using `ReturnType<typeof registerMetrics>` where needed (example in [studio agent](https://github.com/edgeandnode/subgraph-studio/blob/main/packages/agent/src/jobs/settle-invoices.ts#L215))

promaty (approved):

packages/indexer-agent/src/utils/balance-monitor.ts
```diff
@@ -0,0 +1,39 @@
...
+): Promise<void> {
+  logger.info('Monitor operator ETH balance (refreshes every 120s)')
+
+  const balanceMetrics = registerMetrics(metrics)
+
+  timer(120_000).pipe(async () => {
+    try {
+      const balance = await wallet.getBalance()
+      const eth = parseFloat(utils.formatEther(balance))
+      balanceMetrics.operatorEthBalance.set(eth)
```
hopeyen, line 32: could there be an informative log message about how much is the balance? maybe when it is somewhat low?

packages/indexer-agent/src/utils/balance-monitor.ts
Jannis: In fact... I don't even need a type at all here, since TypeScript will infer it from the return type. Good point.

packages/indexer-agent/src/utils/balance-monitor.ts
```diff
@@ -0,0 +1,39 @@
...
+import { indexerError, IndexerErrorCode } from '@graphprotocol/indexer-common'
+
+interface EthBalanceMetrics {
+  operatorEthBalance: Gauge<string>
+}
+
+const registerMetrics = (metrics: Metrics): EthBalanceMetrics => ({
+  operatorEthBalance: new metrics.client.Gauge({
+    name: 'indexer_agent_operator_eth_balance',
+    help: 'Amount of ETH in the operatow wallet; a low amount could cause transactions to fail',
```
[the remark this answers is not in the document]
Jannis: Indeed, good catch!

packages/indexer-agent/src/utils/balance-monitor.ts
Jannis: I've added a log message that includes the ETH balance. What is low depends on a number of factors (e.g. current gas price, average transaction cost) and alerts for what is low may depend from indexer to indexer, who may have different preferences for how early to top up. So I'd prefer to leave that decision to them.

Jannis:
@hopeyen This is ready for another review. :)

hopeyen (approved):
LGTM! thanks Jannis

fordN:
@Jannis Looks like there is an unused import? See the failing `Check formatting` build. 
Should be good to go after that quick fix! ✅

Jannis:
@fordN Oops, yes, removed. I'll merge once CI passes.

PR #357, opened 2022-01-24
Fix indexing status query used for health endpoint

Jannis:
There was an extraneous `{` in the query that caused graph-node to trip over.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,9 @@ name: CI
 
 on:
   push:
-    branches: [master]
+    branches: [main]
   pull_request:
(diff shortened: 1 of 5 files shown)


Jannis:
The CI failures here are unrelated to the PR. It seems that running the indexer test suite requires Postgres but that's not part of the test environment at the moment.

otaviopace (approved):
I'm not very familiar with the project, but all of the changes seem correct 🙂 

If you prefer maybe it's worth waiting for another person take a look as well.

promaty (approved):

fordN (commented):
👍

PR #359, opened 2022-01-28
Use AllocationExchange contract from common-ts

Theodus:
This requires https://github.com/graphprotocol/common-ts/pull/84


Jannis (approved):

fordN:
Looks like we'll need to release a new `common-ts` version to be used here?

Theodus:
> Looks like we'll need to release a new `common-ts` version to be used here?

I think so

PR #361, opened 2022-01-31
indexer-agent: Support deployment id rules using either id format

fordN:
Support both formats of subgraph deployment id: bytes32 or base58.


hopeyen (approved):
looks great! 🥇

PR #362, opened 2022-02-02
indexer-agent reject unsupported subgraph unless bypassed 

hopeyen:
### Goal
- Add a filter such that indexer agent reject to allocate unsupported subgraphs (not eligible for indexing rewards)
- indexers could decide to bypass the filter if desired

### Update
- add additional boolean field `requireSupported` to `IndexingRules`
-- default set to true, such that agent reject unsupported subgraphs by default
-- new rules are set to true by creation
instructions
- if you still want to allocate, bypass the filter by setting `requireSupported` to false
- if you just want to deploy, set `decisionBasis` to offchain


packages/indexer-agent/src/agent.ts
```diff
@@ -11,6 +11,7 @@ import {
 import {
   Allocation,
   allocationRewardsPool,
+  RewardsPool,
```
fordN, line 14: 🤔 I don't see this used below. Is this import needed?

packages/indexer-agent/src/network.ts
```diff
@@ -714,16 +713,27 @@ export class Network {
...
                   },
                 })
 
                 // Skip the indexing rules checks if the decision basis is 'always' or 'never'
                 if (
-                  deploymentRule.decisionBasis === IndexingDecisionBasis.ALWAYS
+                  deploymentRule?.decisionBasis === IndexingDecisionBasis.ALWAYS
                 ) {
+                  // cannot allocate to unsupported subgraph by default
+                  if (
```
fordN, line 725: I think this check should be moved outside of the `deploymentRule?.decisionBasis === IndexingDecisionBasis.ALWAYS` conditional block, so it stops all types of rules from allocating towards `unsupported` subgraphs.

fordN (changes requested):

packages/indexer-agent/src/agent.ts
hopeyen: no idea where it came from 🤨  rebasing and then will remove if still there

hopeyen: I removed it and then found there's a reference to it at line 502 `uniqueRewardsPools: RewardsPool[]`. I don't think it belonged in this PR...

fordN: Ok, now the changlist shows RewardsPool as removed.  Needs to be added back in and in the future if the PRs could be based off main and rebased on main before review that would make it easier to see what changes are made in this PR.

fordN (approved):
👍

PR #364, opened 2022-02-12
Close with latest valid poi for failed subgraphs

hopeyen:
### background

When indexer wants to close an allocation for some failed subgraph with undeterministic errors, the agent receives 0 or null POIs and won't be able to close the allocation. Indexer needed to manually gather and submit POI for failed subgraphs if it fails before the previous epoch. 

### goal
We want the agent to automatically handle POI gathering and closing with the latest valid POI if it exists, and the agent would prefer POI in the order of `current epoch start block >> previous epoch start block >> latest valid block`. 

This feature should reduce manual work and its possibility of gathering the incorrect poi or allocation id which could result in insanely high gas fees.

### updated behavior
when a subgraph encounters a non-deterministic error at certain block,
- failed after `currentEpochStartBlock` —> get poi from `currentEpochStartBlock`
- failed during previous epoch (before `currentEpochStartBlock` and after `previousEpochStartBlock` —> get poi from `previousEpochStartBlock`
- failed before previous epoch —> get poi from the latest valid block (`failureBlock - 1`)
- poi not available —> result in undefined poi OR throw error (agent will not close allocation with zero or null POI

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -806,6 +806,7 @@ class Agent {
             createdAtEpoch: allocation.createdAtEpoch,
             amount: formatGRT(allocation.allocatedTokens),
           })),
+          poi: await this.resolvePOI(epochStartBlock, activeAllocations[0]),
         },
       )
     }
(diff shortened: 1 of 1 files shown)


packages/indexer-agent/src/agent.ts
```diff
@@ -1017,10 +1054,12 @@ class Agent {
       poi === null ||
       poi === utils.hexlify(Array(32).fill(0))
     ) {
+      const tempPoi = await this.resolvePOI(epochStartBlock, existingAllocation)
```
hopeyen, line 1057: should agent be able to reallocate to failed subgraphs? could set poi to this temp poi to reallocate (the current behavior is to keep failed subgraphs open)

packages/indexer-agent/src/agent.ts
```diff
@@ -1010,6 +1041,7 @@ class Agent {
       epochStartBlock,
       this.indexer.indexerAddress,
     )
+    const tempPoi = await this.resolvePOI(epochStartBlock, existingAllocation)
```
fordN, line 1044: I think it's worth reaching out a bit to the indexer community to decide on how this should be handled, but I'd personally want it to work just like close() so it closes with a valid POI even if it has to go back to failed block to do so.

fordN (changes requested):

PR #367, opened 2022-02-21
Use signalledTokens instead of signalAmount for minSignal threshold

Jannis:
The `signalledTokens` field describes the amount of GRT signalled on a subgraph or subgraph deployment. `signalAmount` describes the amount of signal shares on a subgraph, which is based on the bonding curve and hence don't correspond 1:1 to the amount of GRT signalled.

It is therefore much more intuitive to use `signalledTokens` for the `minSignal` threshold in indexing rules.

### DIFF
--- a/packages/indexer-agent/src/network.ts
+++ b/packages/indexer-agent/src/network.ts
@@ -620,7 +620,7 @@ export class Network {
                 id
                 ipfsHash
                 stakedTokens
-                signalAmount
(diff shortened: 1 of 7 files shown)


hopeyen (approved):
The change makes sense to me. re-ran jobs and agent image was able to build :)

abarmat (approved):

PR #370, opened 2022-02-24
indexer-agent: Indexing rules update triggers active allocations query

fordN:
Include indexingRules in the trigger set for activeAllocations updates, so activeAllocations data is up to date when reconciling rules and allocations.


hopeyen (approved):
👍

fordN:
It's not clear this will actually improve the situation, so I'm going to hold off on this until we do some more research.

PR #378, opened 2022-03-07
Indexer-common: Always use base58 format for deployment ids 

fordN:
- Update resolvers to convert subgraph deployment ids to base58 format before saving into the DB. 
- Update eslint and add semi-colon rules.


hopeyen (approved):
lgtm👍

PR #380, opened 2022-03-09
indexer-cli, indexer-common: earliestblock allow null

hopeyen:
Currently if there is a deployment in postgres db with no start block, indexers cannot query `status`
```
Error fetching status information: [GraphQL] Cannot return null for non-nullable field ChainIndexingStatus.earliestBlock.
```

It makes sense to allow null for earliestBlock so that indexers can still query `status` with these deployments.


fordN (approved):
👍

Issue #85, opened 2020-10-19
Feature: allow set cost variables globally


sharppanda:
It would be beneficial, if we can set cost models and variables globally (the same way as we can now do for rules).

variables (very useful):
Global variables are great if cost models are using global variables like GRT_TO_DAI. Currently we need to set those individually for each subgraph.

cost model (convenient):
It is useful when no specific pricing is created for a submodel.

Issue #271, opened 2021-07-13
Feature request: Reconcile allocations and cost models


chriswessels:
Currently there's no built-in process to ensure that any subgraphs that are allocated to also have a cost model defined.

This makes it unfortunately easy for indexers to do the work of indexing a subgraph, without having the prerequisites in place to actually serve queries for it.

Some ideas:
- Update the existing reconciliation loop to check if cost models are defined for all subgraphs, and log a warning if this is not the case
- Allow a default cost model to be defined which will be served for any subgraph that doesn't have a specific cost model defined

fordN:
Being worked on in #381.

PR #381, opened 2022-03-10
Optional to set global cost model and provide default

hopeyen:
Currently when there isn't a cost model defined for a deployment, the Gateway will use a network constant as a fallback for indexer's cost model to the deployment. 

We would like to allow indexers to define a global instance of cost model and variables. Indexers can use indexer-cli to update their preferred fallback instead of the Gateway constant. Indexer agent will take note of deployments with missing cost models and log when global default cost model is created. 

### Usage

Indexer can use indexer-cli to set a global instance of cost model and variables, just like how they usually use indexer-cli to set deployment specific cost models and variables.

Example
```
graph indexer cost set model global [path]/basic.agora
``` 
will either create or update the global cost model if `basic.agora` contains a valid cost model. 

Furthermore, if a global cost model is not defined, indexer agent will look for a cost model at `default.agora`, so this file is only checked when a global cost model doesn't already exist.


packages/indexer-agent/src/indexer.ts
```diff
@@ -332,6 +335,105 @@ export class Indexer {
...
+        }),
+      })
+      return result.data.costModels
+    } catch (error) {
+      this.logger.error(`Failed to query indexer management API`, { error })
+      throw error
+    }
+  }
+
+  async ensureGlobalCostModel(): Promise<void> {
```
hopeyen, line 374: Not sure if this is a good idea (Should I remove `basic.agora` in `indexer-agent` package?) since now indexer could `set model global __` and `set variable global __` through cli

packages/indexer-agent/src/agent.ts
```diff
@@ -599,6 +622,27 @@ class Agent {
...
+    const costModelsIDs = costModels
+      .filter(model => model.deployment != COST_MODEL_GLOBAL)
+      .map(model => new SubgraphDeploymentID(model.deployment))
+    const deploymentMissingModels = targetAllocations.filter(
+      allocation => !(allocation.bytes32 in costModelsIDs),
+    )
+
+    if (deploymentMissingModels) {
+      this.logger.warn(
+        `The following deployments does not have defined cost models and uses global if it exists`,
```
fordN, line 638: Since using plural `deployments` here it should be `do not`, so: 
```
The following deployments do not have defined cost models and will use the global cost model if it exists
```

packages/indexer-agent/src/indexer.ts
```diff
@@ -332,6 +335,105 @@ export class Indexer {
...
+              }
+            }
+          `,
+        )
+        .toPromise()
+
+      if (result.error) {
+        throw result.error
+      }
+      this.logger.debug('Fetched cost models', {
```
fordN, line 357: I'd either remove this or make it a trace log since this action isn't very important for monitoring what the agent is doing. The case that might need attention already has a log; if there isn't a cost model set for a deployment the indexer is allocated to.

packages/indexer-agent/src/indexer.ts
fordN: Good question, let's discuss next time we meet.

fordN (changes requested):
Looks good just a few minor changes and we should discuss the default global cost model.

fordN:
Resolves #85 & #271

packages/indexer-agent/src/indexer.ts
fordN: I think we can remove the default cost model file right now, and simply provide the easy way for indexers to set their default cost model.  That way we aren't defining a default across all indexers that is dependent on their indexer software version.

fordN (approved):
LGTM 👍

packages/indexer-agent/src/indexer.ts
```diff
@@ -13,7 +13,11 @@ import {
   IndexingStatus,
   SubgraphIdentifierType,
   parseGraphQLIndexingStatus,
+  COST_MODEL_GLOBAL,
+  CostModelAttributes,
 } from '@graphprotocol/indexer-common'
+import pRetry from 'p-retry'
```
fordN, line 19: Need pRetry imported here?

PR #385, opened 2022-03-10
indexer-agent: batch receipts when exceeds gateway limit

hopeyen:
Improvement on redeeming receipts for the voucher to avoid hitting the gateway max body size

Testing with the limit of 25,000 receipts
- if requests are less than the limit, exchange requests for a voucher at `/collect_receipts`
- if requests are greater, incremental build the voucher via the `/partial-vouchers` and `/vouchers`

Have yet to test 😅


packages/indexer-agent/src/commands/start.ts
```diff
@@ -774,7 +774,7 @@ export default {
       logger,
       network,
       models: queryFeeModels,
-      collectEndpoint: new URL(argv.collectReceiptsEndpoint),
+      collectEndpoint: argv.collectReceiptsEndpoint,
```
hopeyen, line 777: Would it make sense to use a more general collectEndpoint? or if we could use `--network-subgraph-endpoint` to build `collectEndpoint`, `partialVoucherEndpoint`, and `voucherEndpoint`?

fordN: I think it's ok to keep the separate specific endpoints for now.

packages/indexer-agent/src/query-fees/allocations.ts
```diff
@@ -31,11 +31,18 @@ interface AllocationReceiptsBatch {
   timeout: number
 }
 
+interface PartialVoucher {
```
fordN, line 34: I believe this should include the allocation ID as well. 

```typescript
interface PartialVoucher {
  allocation: string // (0x-prefixed hex)
  fees: string // (0x-prefixed hex)
  signature: string // (0x-prefixed hex)
  receipt_id_min: string // (0x-prefixed hex)
  receipt_id_max: string // (0x-prefixed hex)
}
```

packages/indexer-agent/src/query-fees/allocations.ts
```diff
@@ -300,38 +316,88 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+          )
+          const encodedReceipts = this.encodeReceiptBatch(partial_receipts)
+
+          // Exchange the receipts for a partial voucher signed by the counterparty (aka the client)
+          response = await axios.post(
+            this.partialVoucherEndpoint.toString(),
+            encodedReceipts.unwrap().buffer,
+            { headers: { 'Content-Type': 'application/octet-stream' } },
+          )
+          const partial_voucher = response.data as PartialVoucher
```
fordN, line 377: I think you'll want to parse the response here or simply trim the allocation id hex from the beginning of it.

fordN: The response should have the following structure: 
```typescript
interface PartialVoucher { 
  allocation: string // (0x-prefixed hex) 
  fees: string // (0x-prefixed hex) 
  signature: string // (0x-prefixed hex) 
  receipt_id_min: string // (0x-prefixed hex) 
  receipt_id_max: string // (0x-prefixed hex) }
```

packages/indexer-agent/src/query-fees/allocations.ts
```diff
@@ -300,38 +316,88 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+          partial_vouchers.push(partial_voucher)
+        }
+
+        // Take the partial vouchers and request for a full voucher
+        // (not sure if I can stright up pass the array in, but I want to try encoding them first
+        const voucherRequest = new BytesWriter(
```
fordN, line 390: Let's also move this bytes writer to its own function similar to `encodeReceiptBatch()`?

packages/indexer-agent/src/query-fees/allocations.ts
hopeyen: I thought it should have the structure by `as PartialVoucher` at the end since the interface was declared earlier, should I check if the type fails?

hopeyen:
merged to main on another branch

PR #389, opened 2022-03-12
Agent check previous allocation POI and does not open if zero

hopeyen:
Adding a safety check before opening an allocation: if the subgraph deployment wasn't last closed with a zero POI, the agent will skip allocating it, otherwise just proceed as normal.

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -839,6 +839,25 @@ class Agent {
       logger.info(`No active allocation for deployment, creating one now`, {
         allocationAmount: formatGRT(desiredAllocationAmount),
       })
+
+      // Skip allocating if the previous allocation for this deployment was closed with a null or 0x00 PoI
+      const closedAllocation = (
+        await this.network.closedAllocations(deployment)
+      )[0]
+      if (
+        closedAllocation &&
+        closedAllocation.poi === utils.hexlify(Array(32).fill(0))
+      ) {
+        logger.warn(
+          `Skipping allocation to this deployment as the last allocation to it was closed with a zero POI`,
+          {
+            deployment,
+            closedAllocation: closedAllocation.id,
+          },
+        )
+        return
+      }
+
       const allocationsCreated = await this.network.allocate(
         deployment,
         desiredAllocationAmount,
--- a/packages/indexer-agent/src/network.ts
+++ b/packages/indexer-agent/src/network.ts
@@ -857,6 +857,72 @@ export class Network {
     }
   }
 
+  async closedAllocations(
+    subgraphDeploymentId: SubgraphDeploymentID,
+  ): Promise<Allocation[]> {
+    try {
+      const result = await this.networkSubgraph.query(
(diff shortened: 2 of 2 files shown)


packages/indexer-agent/src/agent.ts
```diff
@@ -839,6 +839,25 @@ class Agent {
...
+
+      // skip allocation if the previous allocation to the deployment closes with null or 0x00
+      const previousAllocation = (
+        await this.network.previousAllocations(deployment)
+      )[0]
+      if (
+        previousAllocation &&
+        previousAllocation.poi === utils.hexlify(Array(32).fill(0))
+      ) {
+        logger.info(
```
fordN, line 851: Shall we make this a `warn` log, so it stands out a bit more?

packages/indexer-agent/src/network.ts
```diff
@@ -857,6 +857,72 @@ export class Network {
...
+          query allocations($indexer: String!, $subgraphDeploymentId: String!) {
+            indexer(id: $indexer) {
+              allocations: totalAllocations(
+                where: {
+                  indexer: $indexer
+                  status: Closed
+                  subgraphDeployment: $subgraphDeploymentId
+                }
+                first: 5
+                orderBy: closedAtEpoch
```
fordN, line 875: I know the contracts don't allow more than one allocation in the same epoch for a single deployment, but just to be safer and bit more explicit let's order by `closedAtBlockNumber`?

packages/indexer-agent/src/network.ts
```diff
@@ -857,6 +857,72 @@ export class Network {
     }
   }
 
+  async previousAllocations(
```
fordN, line 860: Maybe `closedAllocations`? To keep consistent language around allocation state.

fordN (changes requested):

packages/indexer-agent/src/agent.ts
```diff
@@ -839,6 +839,25 @@ class Agent {
       logger.info(`No active allocation for deployment, creating one now`, {
         allocationAmount: formatGRT(desiredAllocationAmount),
       })
+
+      // skip allocation if the previous allocation to the deployment closes with null or 0x00
```
fordN, line 843: `// Skip allocating if the previous allocation for this deployment was closed with a null or 0x00 PoI`?

packages/indexer-agent/src/agent.ts
```diff
@@ -839,6 +839,25 @@ class Agent {
...
+      const closedAllocation = (
+        await this.network.closedAllocations(deployment)
+      )[0]
+      if (
+        closedAllocation &&
+        closedAllocation.poi === utils.hexlify(Array(32).fill(0))
+      ) {
+        logger.warn(
+          `Skipping allocation to this deployment as the last allocation to it was closed by zero POI`,
```
fordN, line 852: "Skipping allocation to this deployment as the last allocation to it was closed with a zero POI"

fordN (approved):
LGTM!

PR #390, opened 2022-03-14
bug-fix: parse graphQL of indexing status

hopeyen:
With more checks to the indexing status, we discovered a missing type conversion that leads to the following error. With a more explicit conversion the issue should be resolved.
```
[2022-03-14 17:53:09.657 +0000] WARN (IndexerAgent/1 on c0a257d65da3): Proof of indexing could not be queried
    attempt: 8
    retriesLeft: 3
    err: "[GraphQL] Invalid value provided for argument `blockNumber`: String(\"4663370\")"
```

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -977,9 +977,18 @@ class Agent {
           },
         )
       } else {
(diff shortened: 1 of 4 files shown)


fordN (approved):
Looks good!

PR #395, opened 2022-03-23
Indexer Agent Configuration by File

kaiwetlesen:
This pull requests implements file-based configuration for the indexer agent such that the configuration file supplements base defaults for a cluster. The idea here is that default values for mainnet, rinkeby, etc can be supplied as code in the repository, with more important parameters being injected via environment variables and CLI flags.

The order of precedence is as follows:
1. CLI flags
2. Environment variables
3. Configuration file parameters

F.eks. running indexer-agent with a configuration file containing `ethereum: https://ethereum-node-0:8545` and setting `INDEXER_AGENT_ETHEREUM` to `https://ethereum.hosted.service/accountid/`, the effective setting would be `https://ethereum.hosted.service/accountid/`.


kaiwetlesen:
Closing this because it broke the CI pipeline.

PR #403, opened 2022-04-12
Manual Allocation Management

fordN:
Introduces allocation management functions to the `IndexerManagementClient` GraphQL interface and a corresponding module in the indexer-cli tool for interacting with it. This interface allows indexers and external tools to have direct control over allocation strategy; rather than applying rules to be interpreted by the indexer-agent into allocation management actions the indexer (or external tool interfacing with the GraphQL interface) can send direct requests to: create, close, or reallocate allocations. Taking the manual actions will send a transaction to the protocol contracts and will update the indexing rules, so the agent will be able to manage the allocation moving forward. For example: when closing an allocation manually the agent will update any existing rule for that deployment to be `decisionBasis = offchain` so the agent will not attempt to reopen that allocation.

To better allow for a workflow that only uses the new manual allocation management commands there is now a new startup parameter `--allocation-management` that can be used to choose between `auto` and `manual` allocation management modes.  The `auto` mode will behave as the agent has in the past where it will continuously monitor the network and use the indexing rules to decide on and execute allocation management actions.  The newly introduced `manual` allocation management mode turns the auto actions of the agent off, so it will not take any allocation management actions on it's own and an indexer can take all the allocation management actions manually using the indexer-cli or IndexerManagementClient GraphQL interface.  

## IndexerManagementClient: GraphQL interface additions
The GraphQL client schema has been updated to include an additional query and 3 additional mutations. 

### Query
```
allocations(filter: AllocationFilter!): [Allocation!]!
```

### Mutations
```
createAllocation(deploymentID: String!, amount: String!, indexNode: String): CreateAllocationResult!
closeAllocation(id: String!, poi: String, force: Boolean): CloseAllocationResult!
reallocateAllocation(
      id: String!
      poi: String
      amount: String!
      force: Boolean
    ): reallocateAllocationResult!
```

## Indexer-cli: new allocations module
The indexer-cli now has an `allocations` module for directly managing the indexer's allocations.

### Indexer allocations get...
The `indexer allocations get` command allows the indexer to query their indexer allocations from the CLI. All 
of the `allocations get` commands query the indexer's own allocations except for `allocations get <allocation-id>` which will query for a specific allocation regardless of which indexer created it. 
```
graph-indexer indexer allocations get all
graph-indexer indexer allocations get <allocation-id>
graph-indexer indexer allocations get --status active
graph-indexer indexer allocations get --status closed
graph-indexer indexer allocations get --deployment <deployment-id>
```

### Indexer allocations create...
```
graph-indexer indexer allocations create <deployment-id> <amount>
```

### Indexer allocations close...
```
graph-indexer indexer allocations close <allocation-id> <poi**>. 
```

### Indexer allocations reallocate...
```
graph-indexer indexer allocations create <allocation-id> <amount> <poi**>
```
**: The PoI parameter is optional; if provided it will be compared against a PoI resolved from the graph-node and the request
will be rejected if they do not match.  The user can also specify `--force` (`-f`) to bypass the PoI accuracy check and force it to send the close transaction with the supplied Poi. 

_note for reviewers: This PR is part of a larger effort to restructure the indexer components, moving network actions into indexer-common, so they can be controlled by the IndexerManagementClient.  This PR is not as DRY as it could be; I plan to consolidate the allocation management functions into a single shared module in a subsequent PR where I will be implementing the action queue._


fordN:
I moved a majority of the dependency and config updates to another PR and rebased this onto that work to reduce the scope of changes to review here.

packages/indexer-common/src/indexer-management/resolvers/allocations.ts
```diff
@@ -0,0 +1,1250 @@
...
+        {
+          indexer: address.toLocaleLowerCase(),
+          status: AllocationStatus[AllocationStatus.Active],
+        },
+      )
+
+      if (result.error) {
+        throw result.error
+      }
+      logger.info('Results', { result })
```
hopeyen, line 475: Maybe we can specify the results and lower the level to `trace` or `debug`

hopeyen (approved):
Nice refactoring of lots of different pieces :D `create` and `close` works pretty well on the testnet

packages/indexer-cli/src/commands/indexer/status.ts
```diff
@@ -23,6 +23,7 @@ module.exports = {
   description: 'Check the status of an indexer',
   run: async (toolbox: GluegunToolbox) => {
     const { print } = toolbox
+    print.warning('We about to get statusy!!')
```
chriswessels, line 26: Is this log intended?

packages/indexer-common/src/indexer-management/client.ts
```diff
@@ -222,6 +276,15 @@ const SCHEMA_SDL = gql`
 
     storeDisputes(disputes: [POIDisputeInput!]!): [POIDispute!]
     deleteDisputes(allocationIDs: [String!]!): Int!
+
+    createAllocation(deploymentID: String!, amount: String!): CreateAllocationResult!
+    closeAllocation(id: String!, poi: String, force: Boolean): CloseAllocationResult!
```
chriswessels, line 281: For `closeAllocation` and `reallocateAllocation`, I like that a POI can be specified, but I think that the default behaviour should be that `indexer-agent` attempts to fetch a valid POI from `graph-node`, and if a POI is provided, it's effectively an override.

packages/indexer-common/src/indexer-management/subgraphs.ts
```diff
@@ -0,0 +1,200 @@
...
+  IndexerManagementModels,
+  IndexingDecisionBasis,
+  IndexingRuleAttributes,
+  SubgraphIdentifierType,
+} from '@graphprotocol/indexer-common'
+import { Logger, SubgraphDeploymentID } from '@graphprotocol/common-ts'
+import jayson, { Client as RpcClient } from 'jayson/promise'
+import pTimeout from 'p-timeout'
+
+export class SubraphManager {
```
chriswessels, line 13: Minor Typo

packages/indexer-common/src/indexer-management/client.ts
```diff
@@ -222,6 +276,15 @@ const SCHEMA_SDL = gql`
 
     storeDisputes(disputes: [POIDisputeInput!]!): [POIDispute!]
     deleteDisputes(allocationIDs: [String!]!): Int!
+
+    createAllocation(deploymentID: String!, amount: String!): CreateAllocationResult!
```
chriswessels, line 280: I'd like to see the ability to specify an override for the `index-node` assignment when creating or reallocating. The reason for this is that some subgraphs have unusual performance profiles (which impacts indexing performance of colocated deployments), and it's desirable to allocate these on a specific `index-node`.

packages/indexer-common/src/indexer-management/subgraphs.ts
```diff
@@ -0,0 +1,200 @@
...
+    } catch (error) {
+      const err = indexerError(IndexerErrorCode.IE027, error)
+      logger.error(`Failed to remove subgraph deployment`, {
+        deployment: deployment.display,
+        err,
+      })
+    }
+  }
+
+  async reassign(logger: Logger, deployment: SubgraphDeploymentID): Promise<void> {
```
chriswessels, line 149: Maybe it's worth exposing this as a GraphQL mutation? Obviously one can call the `reassign` rpc directly on a a graph-node, but it might be nice to expose this to make indexer-agent's API comprehensive and avoid calling dependencies directly.

packages/indexer-common/src/transactions.ts
```diff
@@ -0,0 +1,396 @@
...
+  timer,
+  toAddress,
+} from '@graphprotocol/common-ts'
+import delay from 'delay'
+import { IndexerError, indexerError, IndexerErrorCode } from './errors'
+import { TransactionConfig, TransactionType } from './types'
+import { NetworkSubgraph } from './network-subgraph'
+import gql from 'graphql-tag'
+
+export class TransactionManager {
```
chriswessels, line 25: Is this where we expect the Action Queue implementation will hook in?

packages/indexer-service/src/query-fees/allocations.ts
```diff
@@ -86,10 +90,13 @@ export class AllocationReceiptManager implements ReceiptManager {
     }
 
     const receipt = this._parseAllocationReceipt(receiptData)
+    console.log('Parsed receipt: ', receipt)
```
chriswessels, line 93: This should probably be a call to `logger` instead of `console`? Same below on L99.

packages/indexer-agent/src/__tests__/indexer.ts
```diff
@@ -174,7 +177,7 @@ describe('Indexer tests', () => {
   afterEach(teardown)
 
   // test('Parse Dispute from GraphQL', async () => {})
-  test('Store POI Disputes rejects invalid indexer address', async () => {
+  test('Store PoI Disputes rejects invalid indexer address', async () => {
```
chriswessels, line 180: Particularly given that the code still uses `POI` instead of `PoI`, and the fact that `PoI` is more arduous to type than `POI`, I'd prefer sticking with `POI`.

packages/indexer-agent/src/agent.ts
```diff
@@ -441,14 +447,26 @@ class Agent {
...
+            await this.reconcileAllocations(
+              activeAllocations,
+              targetAllocations,
+              indexingRules,
+              currentEpoch.toNumber(),
+              currentEpochStartBlock,
+              maxAllocationEpochs,
+            )
+          } else if (
+            this.allocationManagementMode == AllocationManagementMode.MANUAL
```
chriswessels, line 460: It looks like `this.reconcileAllocations(...)` contains the code path for collecting receipts and exchanging them for a voucher. We need to expose this capability to the operator some how, or handle receipt collection in a separate loop. https://github.com/graphprotocol/indexer/blob/3813def52ab984e5c4d142a2607636eff2a737d4/packages/indexer-agent/src/agent.ts#L1039

packages/indexer-cli/src/__tests__/references/indexer-rules.stdout
```diff
@@ -1,3 +1,11 @@
 Configure indexing rules
 
-
+  indexer rules stop (never)         Never index a deployment (and stop indexing it if necessary)     
+  indexer rules start (always)       Always index a deployment (and start indexing it if necessary)   
+  indexer rules set                  Set one or more indexing rules                                   
+  indexer rules prepare (offchain)   Offchain index a deployment (and start indexing it if necessary)
```
chriswessels, line 6: How does offchain indexing work if the allocation management mode is `manual`?

chriswessels (commented):
Whoop! So excited to see this landing. Thanks @fordN!

packages/indexer-common/src/indexer-management/client.ts
chriswessels: Oh, I see your comment in the PR:

```**: The PoI parameter is optional; if provided it will be compared against a PoI resolved from the graph-node and the request
will be rejected if they do not match. The user can also specify --force (-f) to bypass the PoI accuracy check and force it to send the close transaction with the supplied Poi.
```

Ignore my comment :)

packages/indexer-cli/src/commands/indexer/status.ts
fordN: Haha no 😆 .  I must not have pushed the final once-over I did to remove debug logs.

packages/indexer-cli/src/__tests__/references/indexer-rules.stdout
fordN: In manual mode allocation management is turned off, but deployment management continues as before.  
As a result the behavior is very similar to before for `offchain` subgraphs.   i.e - `offchain` subgraphs will be deployed to the index nodes but allocations will not be created or managed for them.   The difference is if a subgraph with an active allocation is moved to `offchain` the agent will no longer close that allocation for you.

packages/indexer-service/src/query-fees/allocations.ts
fordN: I'll remove these logs.  I actually had them in here to debug something else entirely 😉

packages/indexer-common/src/transactions.ts
fordN: Largely yes; I expect the action queue worker implementation (that picks up action items and executes them) will essentially hook in here.   Additionally there will be graphql resolvers for adding items to the action queue that the indexer-agent (or external tools) will call in place of it's current transaction execution.

packages/indexer-common/src/indexer-management/subgraphs.ts
fordN: Absolutely.  I actually have those resolvers written up and stashed; I decided to table that change for a subsequent PR since they aren't essential to allocation management and I'd like to include a corresponding indexer-cli module to go with them (`indexer subgraphs deploy`, etc..).

packages/indexer-common/src/indexer-management/client.ts
fordN: Good idea, I'll add an optional string parameter for specifying an index-node.

packages/indexer-common/src/indexer-management/resolvers/allocations.ts
fordN: Thanks for pointing this out.  Looks like I left some debug logs in! Oops, I'll remove.

packages/indexer-agent/src/agent.ts
fordN: ok, so right now `collectReceipts()` is coupled with `closingAllocation()` in the agent reconciliation loop. 

I could certainly continue that pattern in the manual allocation management workflow by calling collect receipts when `indexer allocations close..` or `indexer allocations reallocate..` are used. However, this may be a good time to decouple closing from collecting and move `collectReceipts()` into a separate reconciliation loop function that has Eventual<recentlyClosedAllocations> as an input. 

In the case of allocationManagmentMode = manual, which behavior would you personally prefer? 
- A separate loop that uses the current threshold configs to automatically collect receipts for closed allocations. 
- Coupling collectReceipts with `allocations close` and `allocations reallocate` so the manual allocation actions lead to automatic managing of receipts. 
- Separate out collecting of receipts to a separate indexer-cli command, so indexers have to call it explicitly to collect receipts for an allocation.  If we did this I'd also include a command to get a summary of receipts per allocation to be a part of this workflow. Another option with this more manual workflow is to use a separate setting to control receipt collection; then we'd have both the `allocation-management` config and a separate config like `receipt-management` which can also be set to `auto` or `manual`.

packages/indexer-agent/src/__tests__/indexer.ts
fordN: I've been waffling a bit back and forth on this; happy to stick to `POI`.

packages/indexer-agent/src/agent.ts
fordN: I think for this PR I'm going to translate the current close allocation behavior into the manual action, so I'll add the `collectReceipts()` execution at the end of the manual close allocation process.  
We can continue this conversation about how to move forward with the design and I'll implement the output of that discussion in a subsequent PR, likely with the action queue work.

fordN:
Thanks for the reviews and suggestions @hopeyen, @chriswessels, and @kaiwetlesen ! 🙏 

Here are the updates I'm making thanks to your input.  I'll update this list in place as I check them off. Please let me know if I missed anything!

- [X] Add optional `indexNode` parameter to `createAllocation` for specifying the node that subgraph should be deployed to.
- [X] Use consistent formatting for POI (`POI` instead of `PoI`).
- [X] Include `collectReceipts` in the manual `closeAllocation` process.
- [X] Remove debug logs I should have already remove ;). 
- [ ] Support returning manual allocation CLI command result objects in YAML, JSON, and table format.

PR #404, opened 2022-04-13
*: Upgrade dependencies, cleanup unused, update prettier config

fordN:


lerna.json
```diff
@@ -2,5 +2,7 @@
   "packages": [
     "packages/*"
   ],
+  "npmClient": "yarn",
+  "useWorkspaces": true,
```
kaiwetlesen, line 6: Will this fix the issue with running `npm install`?

package.json
```diff
@@ -25,15 +25,19 @@
...
     "@ethersproject/bignumber": "5.5.0",
     "@ethersproject/providers": "5.5.1",
-    "@urql/core": "2.3.6",
-    "@urql/exchange-execute": "1.2.2",
-    "scrypt": "https://registry.yarnpkg.com/@favware/skip-dependency/-/skip-dependency-1.0.2.tgz"
+    "@urql/core": "2.4.4",
+    "@urql/exchange-execute": "1.2.2"
   },
-  "dependencies": {
-    "scrypt": "https://registry.yarnpkg.com/@favware/skip-dependency/-/skip-dependency-1.0.2.tgz"
```
kaiwetlesen, line 36: Did we ever determine the reason for this dependency skip?

packages/indexer-agent/package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "format": "prettier --write src/*.ts src/**/*.ts src/**/**/*.ts",
     "lint": "eslint . --ext .ts,.tsx",
-    "compile": "./node_modules/.bin/tsc",
+    "compile": "tsc",
```
kaiwetlesen, line 13: Will this work under all platforms wherein the path doesn't always get set up correctly?

packages/indexer-cli/package.json
```diff
@@ -7,41 +7,41 @@
   "author": "The Graph Foundation",
   "license": "MIT",
   "scripts": {
-    "format": "prettier --write src/*.ts src/**/*.ts src/**/**/*.ts",
-    "lint": "eslint . --ext .ts,.tsx",
-    "compile": "./node_modules/.bin/tsc",
-    "prepare": "rm -rf dist && yarn format && yarn lint && yarn compile",
+    "format": "prettier --write 'src/**/*.ts'",
+    "lint": "eslint . --ext .ts,.tsx --fix",
```
kaiwetlesen, line 11: 🚀

packages/indexer-cli/src/commands/indexer/cost/get.ts
```diff
@@ -50,15 +50,16 @@ module.exports = {
       : ['deployment', 'model', 'variables']
     const rawDeployment = ['model', 'variables'].includes(first) ? second : first
 
-    try{
-      if (rawDeployment !== 'all') {await validateDeploymentID(rawDeployment)}
+    try {
+      if (rawDeployment !== 'all') {
+        await validateDeploymentID(rawDeployment)
+      }
```
kaiwetlesen, line 56: Guess this is the linter fix?

lerna.json
fordN: Unfortunately doesn't fully fix it, but is part of my work to get there.  It's at least setting this mono repo up in a more explicit way.

package.json
fordN: No, I couldn't find it being used anywhere so I simply removed.  I'm inclined to think that it was used for vector/connext payment channels before they were removed.

kaiwetlesen (approved):
Looks good! Guess there's just a couple questions we should have written down somewhere.

packages/indexer-agent/package.json
fordN: Good question.  I don't have a definitive answer, but found that this makes it work with NPM whereas the node_modules/bin location execution we were using before did not.

packages/indexer-cli/src/commands/indexer/cost/get.ts
fordN: Yep!

lerna.json
fordN: I need to do some more research on handling workspaces in `npm` and generally setting up a monorepo to work with both `npm` and `yarn.

fordN:
Thanks for the quick review @kaiwetlesen! 🙏   
I'll merge when the CI builds and tests complete successfully.

Issue #387, opened 2022-03-11
Do not hold onto receipts with 0 value


fordN:
Towards resolving #278  

The contracts do not support collecting receipts with 0 value, so let's disregard 0 value receipts early on instead of saving them. I'm expecting the update to take place in the [AllocationReceiptManager](https://github.com/graphprotocol/indexer/blob/main/packages/indexer-service/src/query-fees/allocations.ts#L82).

PR #405, opened 2022-04-14
indexer-service: filter 0 value receipts

hopeyen:
https://github.com/graphprotocol/indexer/issues/387

Disregard 0 value receipts early on instead of saving them as they would won't be collected by the contracts anyway.


fordN (approved):
👍

PR #407, opened 2022-04-19
WIP: Use ECMAScript modules 

fordN:
Towards supporting building the indexer component using either CommonJS or ECMAScript module (ESM) loading. Ideally we'd support both in order to provide compatibility with a larger set of the node ecosystems libraries, npm/yarn versions, and machine types. 

Some background: 
- Projects that use ESM can easily load CommonJS libraries but not the other way around.  Since some of the libraries we depend on have updated to ESM we are sticking to older CommonJS versions of them leaving us potentially vulnerable to security issues or bugs that have since been fixed.
- ESM allows for more control of imports and if used well could lead to a significant decrease in size of the distributed packages. 
- ESM has a few other benefits such as asynchronous loading, top-level await, and improved static analysis support.


dwerner:
Closing this as it's a bit stale.

PR #421, opened 2022-05-13
indexer-cli: indexing rule parser error catch

hopeyen:
When using wrong arguments for `indexer rules set`, the error message 
`TypeError: INDEXING_RULE_PARSERS[key] is not a function` 
is not very clear. 

Improve the error message to specify the invalid key: when provided a non-existing attribute, return messages like
`Indexing rule attribute 'allocatoinAmount' doesn't exist, please check spelling`

for next step: can make suggestions using modules like `spellchecker-cli`

### DIFF
--- a/packages/indexer-cli/src/commands/indexer/rules/set.ts
+++ b/packages/indexer-cli/src/commands/indexer/rules/set.ts
@@ -4,9 +4,15 @@ import { partition } from '@thi.ng/iterators'
 
 import { loadValidatedConfig } from '../../../config'
 import { createIndexerManagementClient } from '../../../client'
-import { fixParameters } from '../../../command-helpers'
-import { parseIndexingRule, setIndexingRule, printIndexingRules } from '../../../rules'
+import { fixParameters, suggestCommands } from '../../../command-helpers'
+import {
+  parseIndexingRule,
+  setIndexingRule,
+  printIndexingRules,
+  indexingRule,
+} from '../../../rules'
 import { processIdentifier } from '@graphprotocol/indexer-common'
+import { stringify } from 'yaml'
 
 const HELP = `
 ${chalk.bold('graph indexer rules set')} [options] global          <key1> <value1> ...
@@ -46,6 +52,7 @@ module.exports = {
         global: true,
       })
       const config = loadValidatedConfig()
+      const client = await createIndexerManagementClient({ url: config.api })
 
       //// Parse key/value pairs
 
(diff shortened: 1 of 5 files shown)


fordN:
Nice! Can we also output the help text after the message so the user can see what the supported parameters are?

hopeyen:
> Nice! Can we also output the help text after the message so the user can see what the supported parameters are?

I ended up making a simple dumb suggestion function: if there's valid commands that shares 80% of the inputted letters, print those, otherwise print all available commands. Let me know if you think the help text would be better?

packages/indexer-common/src/indexer-management/models/indexing-rule.ts
```diff
@@ -30,6 +30,24 @@ export interface IndexingRuleAttributes {
   requireSupported: boolean
 }
 
+export const IndexingRuleAttributeDictionary = [
```
fordN, line 33: I don't expect the indexing rule models to change, but in order to ease management and make the code more "DRY" can we avoid creating a new constant here and instant derive the array of keys from the `IndexingRuleAttributes` interface or the `IndexingRule` class?

packages/indexer-cli/src/rules.ts
```diff
@@ -137,7 +137,14 @@ export const parseIndexingRule = (
   const obj = {} as any
   for (const [key, value] of Object.entries(rule)) {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    obj[key] = (INDEXING_RULE_PARSERS as any)[key](value)
+    const parser = (INDEXING_RULE_PARSERS as any)[key]
+    if (!parser) {
+      throw (
+        `Indexing rule attribute '${key}' doesn't exist, please check spelling, might be one of these:\n` +
```
fordN, line 143: What do you think about changing the text to this? 
`Indexing rule attribute '${key}' not supported, did you mean?`

packages/indexer-common/src/indexer-management/models/indexing-rule.ts
hopeyen: I totally agree! 'Instant derive' is what I was missing. Do you think we could use module `ts-transformer-keys`?

fordN: I think this should be doable without any additional libraries.  

Perhaps we can use the rawAttributes property of the Sequelize model?
`Object.keys(IndexingRule.rawAttributes)`?

or use the keys from an instance of the class?
`Object.keys(new IndexingRule())`?

fordN:
@hopeyen what's the status on this PR? Ready for another review?

hopeyen:
> 

yep this one is ready for another review! thanks in advance

packages/indexer-cli/src/commands/indexer/rules/set.ts
```diff
@@ -61,23 +63,31 @@ module.exports = {
...
+            'Parallel allocations are soon to be fully deprecated. Please set parallel allocations to 1 for all your indexing rules',
+          )
+          process.exitCode = 1
+        }
+        const rule = await setIndexingRule(client, inputRule)
+        printIndexingRules(print, outputFormat, identifier, rule, [])
+      } catch (error) {
+        // Failed to parse input, make suggestions
+        const valid_commands = Object.keys(
+          (await indexingRules(client,false))[0],
```
hopeyen, line 84: or I could call `await indexingRule(client,'global', false)`? although that returns type potentially null that needs more conditionals to handle. Getting index 0 here is basically the global rule which is ensured to exist. Not sure if we need to worry about not having an empty list

fordN (approved):
LGTM 👍

PR #422, opened 2022-05-13
indexer-common: start block query - ensure numeric type

hopeyen:
small catch for the potential BlockPointers being returned as string that isn't being parsed to types properly using https://github.com/graphprotocol/indexer/blob/06f0c4465aa4ead000a6ec3db94b44ff551a0a94/packages/indexer-common/src/indexing-status.ts#L35-L38


fordN (approved):
👍

PR #423, opened 2022-05-17
indexer-cli: remove colors from yaml and json ouputs

hopeyen:
To keep consistent between writing in the terminal and file, we remove all the colors codes when someone is outputting JSON or YAML


fordN (approved):
LGTM! I think it just needs a rebase on origin/main.

PR #424, opened 2022-05-19
Dockerfiles: Update base images to node:16.15-bullseye-slim

fordN:
`node:16.13.1-slim` images use an older version of GLIBC not compatible with the
precompiled native binaries. 

This fixes docker image builds, but we'll need to find a solution so that users can build the packages on older versions of linux with older GLIBC versions.

### DIFF
--- a/Dockerfile.indexer-agent
+++ b/Dockerfile.indexer-agent
@@ -1,7 +1,7 @@
 ########################################################################
 # Build image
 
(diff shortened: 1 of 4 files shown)


kaiwetlesen (approved):
This should do it, I was able to build the image and resolve the glibc version mismatch.

PR #431, opened 2022-06-28
Actions Queue: add auto batch options and conditions

hopeyen:
When an indexer uses AUTO management mode, the agent automatically approve actions for the worker to execute. 

To avoid worker executing those transactions right away, we provide `indexer-agent` optional start up parameters:
- `--auto-min-batch-size`: require a minimum batch size 
- `--auto-max-wait-time`: wait for certain minutes after the first approved actions' update time

These 2 parameters are useful when an indexer let the agent perform automatic batched transactions, allow time for indexers to modify actions, and in particular save gas costs on allocations. 

Another condition we added is that actions will still execute right away when an on-chain transactions is expiring - the agent automatically produce and approve reallocation actions and it seems sensible to allow worker to execute them as quickly as possible.

Action queue worker will continue as usual if any of the 3 conditions are met. 

Additionally, within a batch, actions of type `unallocate` are ordered before of type `reallocate` which is before `allocate`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,8 +185,11 @@ Indexer Infrastructure
                                         block gas limit  [number] [default: 100]
   --log-level                           Log level    [string] [default: "debug"]
   --allocation-management               Indexer agent allocation management
-                                        automation mode (auto|manual)
+                                        automation mode (auto|manual|oversight)
                                                       [string] [default: "auto"]
+  --auto-allocation-min-batch-size                 Minimum number of allocation 
+                                        transactions inside a batch for AUTO 
+                                        management mode    [number] [default: 1]
 
 Network Subgraph
   --network-subgraph-deployment   Network subgraph deployment           [string]
--- a/packages/indexer-agent/README.md
+++ b/packages/indexer-agent/README.md
@@ -38,6 +38,12 @@ Indexer Infrastructure
                                 'false' rewards will be returned to the wallet
                                                        [boolean] [default: true]
   --log-level                   Log level            [string] [default: "debug"]
+  --allocation-management               Indexer agent allocation management
+                                        automation mode (auto|manual|oversight)
+                                                      [string] [default: "auto"]
+  --auto-allocation-min-batch-size                 Minimum number of allocation 
+                                        transactions inside a batch for auto 
+                                        management mode    [number] [default: 1]
 
 Network Subgraph
   --network-subgraph-deployment  Network subgraph deployment            [string]
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -421,6 +421,12 @@ export default {
         default: 'auto',
         group: 'Indexer Infrastructure',
       })
+      .option('auto-allocation-min-batch-size', {
+        description: `Minimum number of allocation transactions inside a batch for auto allocation management. No obvious upperbound, with default of 1`,
+        type: 'number',
+        default: 1,
+        group: 'Indexer Infrastructure',
+      })
       .config({
         key: 'config-file',
         description: 'Indexer agent configuration file (YAML format)',
@@ -759,6 +765,10 @@ export default {
     )
 
     logger.info('Launch indexer management API server')
+    const allocationManagementMode =
+      AllocationManagementMode[
+        argv.allocationManagement.toUpperCase() as keyof typeof AllocationManagementMode
+      ]
     const indexerManagementClient = await createIndexerManagementClient({
       models: managementModels,
       address: indexerAddress,
@@ -779,6 +789,8 @@ export default {
       },
       transactionManager: network.transactionManager,
       receiptCollector,
+      allocationManagementMode,
+      autoAllocationMinBatchSize: argv.autoAllocationMinBatchSize,
     })
 
     await createIndexerManagementServer({
@@ -845,10 +857,7 @@ export default {
         (s: string) => new SubgraphDeploymentID(s),
       ),
       receiptCollector,
-      allocationManagementMode:
-        AllocationManagementMode[
-          argv.allocationManagement.toUpperCase() as keyof typeof AllocationManagementMode
-        ],
+      allocationManagementMode,
     })
   },
 }
--- a/packages/indexer-common/src/indexer-management/actions.ts
+++ b/packages/indexer-common/src/indexer-management/actions.ts
@@ -2,8 +2,11 @@ import {
   Action,
   ActionFilter,
   ActionStatus,
+  AllocationManagementMode,
+  AllocationStatus,
   IndexerManagementModels,
   isActionFailure,
+  NetworkMonitor,
 } from '@graphprotocol/indexer-common'
 import { AllocationManager } from './allocations'
 import { Transaction } from 'sequelize'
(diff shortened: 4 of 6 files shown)


packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -12,13 +15,61 @@ import { Eventual, join, Logger, timer } from '@graphprotocol/common-ts'
 export class ActionManager {
   constructor(
     public allocationManager: AllocationManager,
+    public networkMonitor: NetworkMonitor,
+    private currentEpoch: number,
+    private maxAllocationEpoch: number,
```
hopeyen, line 20: some alternative to passing in these parameters I was thinking were we could pass in `network: Network` and call contract functions, or implement similar `network.contracts` functions?

packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -12,13 +15,61 @@ import { Eventual, join, Logger, timer } from '@graphprotocol/common-ts'
...
+
+    // In AUTO mode, worker waits to execute when 1) the batch is not filled upto minimum size
+    // and 2) oldest allocatation is not expiring after the current epoch
+    // and 3) oldest action is not stale (10 times the user provided threshold of polling frequency)
+    if (this.allocationManagementMode === AllocationManagementMode.AUTO) {
+      const oldestCreatedAtEpoch = (
+        await this.networkMonitor.allocations(AllocationStatus.ACTIVE)
+      )[0].createdAtEpoch
+
+      const updatedDeadline = new Date(new Date().valueOf() - 10 * pollingFrequency)
```
hopeyen, line 44: Not sure if this relationship makes sense with L66 and 68. I didn't want to add another start-up parameter, but I can see how the 2 time thresholds shouldn't be related

packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -49,11 +104,20 @@ export class ActionManager {
...
-          where: { status: ActionStatus.APPROVED },
-          transaction,
-          lock: transaction.LOCK.UPDATE,
+        // Execute already approved actions in the order of type and priority
+        const actionTypePriority = ['unallocate', 'reallocate', 'allocate']
+        const approvedActions = (
+          await this.models.Action.findAll({
+            where: { status: ActionStatus.APPROVED },
+            order: [
+              // ['type', 'DESC'],
```
hopeyen, line 113: This returns actions types in order of `reallocate`, `unallocate`, and then `allocate` for some reason? `ASC` returns the reversed. I sorted afterwards on L119 instead, but might be nice if I can do a sequelize ordering

packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -12,13 +15,61 @@ import { Eventual, join, Logger, timer } from '@graphprotocol/common-ts'
...
 
+  private async batchReady(
+    approvedActions: Action[],
+    pollingFrequency: number,
+  ): Promise<boolean> {
+    if (approvedActions.length < 1) {
+      return false
+    }
+
+    // In AUTO mode, worker waits to execute when 1) the batch is not filled upto minimum size
```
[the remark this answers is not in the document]
fordN: I'd be for removing the `approvedActionsStale` condition though.  If the batch size doesn't meet the minimum the indexer can use the `indexer actions execute approved` command to force execution? 

Do you think that is enough? Or will indexers want some safety mechanism like the approved item has been waiting too long condition you propose in this PR?

packages/indexer-agent/src/commands/start.ts
```diff
@@ -421,6 +421,18 @@ export default {
         default: 'auto',
         group: 'Indexer Infrastructure',
       })
+      .option('auto-min-batch-size', {
+        description: `Minimum number of allocation transactions inside a batch for AUTO allocation management. No obvious upperbound, with default of 1`,
+        type: 'number',
+        default: 1,
+        group: 'Indexer Infrastructure',
+      })
+      .option('auto-max-wait-time', {
```
fordN, line 430: Maybe if this isn't defined by the user it is infinite or 0? So in that case approved actions will auto execute only if the batch size meets the `auto-min-batch-size`?

packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -12,13 +15,61 @@ import { Eventual, join, Logger, timer } from '@graphprotocol/common-ts'
...
+  }
+
   async monitorQueue(): Promise<void> {
-    const approvedActions: Eventual<Action[]> = timer(30_000).tryMap(
-      async () => await this.fetchActions({ status: ActionStatus.APPROVED }),
+    const pollingFrequency = 1000 * 60 * (this.autoMaxWaitTime ? this.autoMaxWaitTime : 1)
+
+    const approvedActions: Eventual<Action[]> = timer(pollingFrequency).tryMap(
+      async () =>
+        await this.models.Action.findAll({
```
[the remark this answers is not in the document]
hopeyen: Oh, you're right! That's a better way, thanks.

packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -12,13 +15,61 @@ import { Eventual, join, Logger, timer } from '@graphprotocol/common-ts'
...
+    // and 2) oldest allocatation is not expiring after the current epoch
+    // and 3) oldest action is not stale (10 times the user provided threshold of polling frequency)
+    if (this.allocationManagementMode === AllocationManagementMode.AUTO) {
+      const oldestCreatedAtEpoch = (
+        await this.networkMonitor.allocations(AllocationStatus.ACTIVE)
+      )[0].createdAtEpoch
+
+      const updatedDeadline = new Date(new Date().valueOf() - 10 * pollingFrequency)
+
+      this.logger.debug('Worker in AUTO, checking batch conditions', {
```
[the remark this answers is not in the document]
hopeyen: There's an info log `Executing batch of approved actions` for when `batchReady` is true, and the conditional logging at line92 is the complement of that. I guess they were mainly for me to check the results quickly, but I could remove both logs if `Executing batch of approved actions` is enough?

packages/indexer-common/src/indexer-management/actions.ts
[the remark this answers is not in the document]
hopeyen: I am in favor of removing the `approvedActionsStale` condition, that means we don't need the additional indexer agent startup parameter `auto-max-wait-time`. I think they can use the `execute` command if they are in a rush:D

packages/indexer-common/src/indexer-management/actions.ts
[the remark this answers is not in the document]
fordN: Yeah, I think `Executing batch of approved actions` is enough as long as it includes all the info. Do you think having only that would make it unclear what's happening in the logs?

packages/indexer-common/src/indexer-management/actions.ts
fordN: I think it's a good idea to implement functions to get epoch config data in `NetworkMonitor`.

packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -12,10 +15,43 @@ import { Eventual, join, Logger, timer } from '@graphprotocol/common-ts'
...
   constructor(
     public allocationManager: AllocationManager,
+    public networkMonitor: NetworkMonitor,
+    private currentEpoch: number,
+    private maxAllocationEpoch: number,
     private logger: Logger,
     private models: IndexerManagementModels,
+    private allocationManagementMode?: AllocationManagementMode,
+    private autoMinBatchSize?: number,
+    private autoMaxWaitTime?: number,
```
fordN, line 25: Looks like this value is no longer being used. (?)

Let's remove? The `affectedAllocationExpiring` check should be sufficient for forcing batch executing before the `autoMinBatchSize` is met.

packages/indexer-agent/src/commands/start.ts
fordN: Do you think the name is immediately clear? `auto-min-batch-size` 

I'm not sure that tells a user what kind of batch size it is (we have many batching mechanisms 😄  ). 

What about `auto-allocation-min-batch-size` ?

packages/indexer-agent/src/commands/start.ts
```diff
@@ -779,6 +797,9 @@ export default {
       },
       transactionManager: network.transactionManager,
       receiptCollector,
+      allocationManagementMode,
+      allocationAUTOMinBatchSize: argv.autoMinBatchSize,
```
fordN, line 801: Why the AUTO in call caps?  Let's use camel case? `autoAllocationMinBatchSize`?

packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -49,11 +85,17 @@ export class ActionManager {
     await this.models.Action.sequelize!.transaction(
       { isolationLevel: Transaction.ISOLATION_LEVELS.SERIALIZABLE },
       async (transaction) => {
-        // Execute already approved actions first
-        const approvedActions = await this.models.Action.findAll({
-          where: { status: ActionStatus.APPROVED },
-          transaction,
-          lock: transaction.LOCK.UPDATE,
+        // Execute already approved actions in the order of type and priority
+        const actionTypePriority = ['unallocate', 'reallocate', 'allocate']
```
fordN, line 89: Can we include a quick explanation of the type ordering? i.e why "unallocate" first?

packages/indexer-common/src/indexer-management/actions.ts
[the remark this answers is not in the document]
hopeyen: I think it might be tricky to include all the info in `Executing batch of approved actions` as it lives outside of `batchReady`, thinking about this again, I'll add a log after conditional tests like you suggested that shows the results:)

packages/indexer-agent/src/commands/start.ts
hopeyen: very true, concise names are hard to think of sometimes!

packages/indexer-common/src/indexer-management/actions.ts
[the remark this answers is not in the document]
fordN: ok!

packages/indexer-agent/src/commands/start.ts
fordN: Definitely!

packages/indexer-agent/README.md
```diff
@@ -38,6 +38,12 @@ Indexer Infrastructure
                                 'false' rewards will be returned to the wallet
                                                        [boolean] [default: true]
   --log-level                   Log level            [string] [default: "debug"]
+  --allocation-management               Indexer agent allocation management
+                                        automation mode (auto|manual|oversight)
```
fordN, line 42: nice, good eye.

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -80,7 +87,11 @@ export class NetworkMonitor {
       const result = await this.networkSubgraph.query(
         gql`
           query allocations($indexer: String!, $status: AllocationStatus!) {
-            allocations(where: { indexer: $indexer, status: $status }, first: 1000) {
+            allocations(
+              where: { indexer: $indexer, status: $status }
+              first: 1000
+              orderBy: createdAtBlockNumber
+            ) {
```
fordN, line 94: Shall we be explicit here and specify the orderDirection?

README.md
```diff
@@ -185,8 +185,11 @@ Indexer Infrastructure
                                         block gas limit  [number] [default: 100]
   --log-level                           Log level    [string] [default: "debug"]
   --allocation-management               Indexer agent allocation management
-                                        automation mode (auto|manual)
+                                        automation mode (auto|manual|oversight)
```
fordN, line 188: Nice, good eye.

fordN (approved):
LGTM! 👍 

I suggested one change, but it doesn't affect the runtime it's just for improved clarity.

packages/indexer-common/src/indexer-management/monitor.ts
hopeyen: added `orderDirection: asc` for oldest ones first! Thanks for the suggestion

PR #432, opened 2022-06-29
Indexer-cli: only print identifier for offchain subgraphs

hopeyen:
For offchain subgraphs, indexing rules fields like `allocationAmount` doesn't really matter, so the table view is more than enough. For a clearer output, we simply print out offchain subgraphs' list of identifiers.


fordN (commented):
Are there other places that we display offchain syncing subgraphs that should also be updated similarly? `indexer status`?

hopeyen:
> Are there other places that we display offchain syncing subgraphs that should also be updated similarly? `indexer status`?

I think `indexer status` is going to call `printIndexingRules` here and follow the same format there, so it should be taken care of. I couldn't think of anywhere else that prints offchain subgraphs:)

fordN (approved):
LGTM! 👍

PR #446, opened 2022-08-02
docs/network: Update recommended graph-node version

evaporei:
https://github.com/graphprotocol/graph-node/releases/tag/v0.27.0


fordN:
@evaporei there was some discussion of indexers having issues with this version.  Is it stable for indexer use? If there are specific issues being worked on that we can link to here?

evaporei:
Hey @fordN both issues are now solved, no need for a new patch (eg 0.27.1).
This can now be recommended for indexers 🙂 
cc @azf20

azf20 (approved):

Issue #443, opened 2022-07-27
Bug: indexer status fails if a chain indexing status is missing latestBlock value


fordN:
Error: 
```shell
./bin/graph-indexer indexer status                     
Error fetching status information: [GraphQL] Cannot return null for non-nullable field ChainIndexingStatus.latestBlock.
[GraphQL] Cannot return null for non-nullable field ChainIndexingStatus.latestBlock.
```

PR #453, opened 2022-08-03
indexer-cli, -common: allow status latestBlock to be nullable

hopeyen:
for https://github.com/graphprotocol/indexer/issues/443
A few months ago we made `earliestBlock` nullable which is a common occurrence for subgraph deployments. For unlikely edge cases we allow `latestBlock` and `chainHeadBlock` to be null as well.


packages/indexer-common/src/indexer-management/__tests__/actions.ts
```diff
@@ -245,7 +245,7 @@ const setup = async () => {
   })
   networkSubgraph = await NetworkSubgraph.create({
     logger,
-    endpoint: 'https://gateway.testnet.thegraph.com/network',
+    endpoint: 'https://gateway.thegraph.com/network',
```
fordN, line 248: I've updated all these to use hosted-service endpoints recently; let's do that here too?  If you rebase on main you should see it come through.

fordN (approved):
Looks good! Approved to merge after you rebase on main and use the testnet network subgraph hosted-service endpoint in the tests.

fordN:
Resolves #443

PR #454, opened 2022-08-08
Shorten failure reason stored

### LINKED ISSUE #444: Action queue improvements: Improve failure reason storage
Currently the failure reason column gets populated by the error log message if a transaction fails.  These log messages were designed to be logged to stdout _not_ to be stored in a DB column. We should update the failure reason stringsso they are short, to the point, well formatted failureReason strings ready to be stored in the DB.

hopeyen:
Resolves #444 

Currently the failure reason column gets populated by the error log message if a transaction fails. These log messages were designed to be logged to stdout not to be stored in a DB column. Occasionally the messages seem to hit the limit `SequelizeDatabaseError: value too long for type character varying(1000)`.  

We update the failure reasons to be shorter strings. For the actions that failed due to an existing `IndexerError`, we simply store the error code as `IE0XX` so people can look up the reasons and solutions if provided. We also added more variations of IndexerError to cover the failure cases here with the action manager.

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -608,7 +608,7 @@ See also: [#IE041](#ie041).
 
 **Summary**
 
-TODO
+Failed to collect query fees on chain.
 
 **Solution**
 
@@ -618,7 +618,7 @@ TODO
 
 **Summary**
 
-TODO
+Failed to queue transfers for resolving.
 
 **Solution**
 
@@ -628,7 +628,7 @@ TODO
 
 **Summary**
 
-TODO
+Failed to resolve transfer.
 
 **Solution**
 
@@ -638,7 +638,7 @@ TODO
 
 **Summary**
 
-TODO
+Failed to mark transfer as failed.
 
 **Solution**
 
@@ -648,7 +648,7 @@ TODO
 
 **Summary**
 
-TODO
+Failed to withdraw query fees for allocation.
 
 **Solution**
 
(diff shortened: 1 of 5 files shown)


docs/errors.md
```diff
@@ -658,8 +658,200 @@ TODO
...
+
+## IE066
+
+**Summary**
+
+Allocation not created: Allocation ID already exist.
+
+**Solution**
+
+Cannot create a new allocation. The created allocation ID already exist on-chain. Check the network sync status, if status is caught up and an allocation still needs to be opened, agent will create a new unique allocation ID and try again.
```
fordN, line 835: ```Duplicate allocation ids are not permitted. Check the network subgraph sync status, if it is caught up and an allocation still needs to be opened, agent will create a new unique allocation ID and try again.```

docs/errors.md
```diff
@@ -658,8 +658,200 @@ TODO
...
+
+## IE067
+
+**Summary**
+
+Failed to resolve POI: POI not available for deployment at current epoch start block.
+
+**Solution**
+
+Check sync and health status of the subgraph to access the issue. Might need to wait index node the sync up and is able to generate the current epoch start block POI.
```
fordN, line 845:
```
This error typically indicates a subgraph deployment that is too far behind the chain head to resolve a valid POI for closing the allocation. To succesfully resolve the POI and close the allocation you'll need to wait for the index node to sync the deployment enough to generate a POI for the current epoch start block. If you do not want to wait to sync you may choose to forfeit indexing rewards for the allocation by force closing it with a 0 POI by creating an unallocate or reallocate action item: `graph indexer actions queue unallocate/reallocate <deployment-id> <allocation-id> 0 true`.
```

fordN (approved):
LGTM 👍

PR #457, opened 2022-08-11
indexer-cli,-common: ordering action queries

### LINKED ISSUE #442: Improvement: better default ordering of `indexer actions get` results, and provide user control
By default we should provide a clear ordering pattern. I'd propose ordering by id, asc. 

I would also like to add an optional ordering control parameter... or two.  We could use a similar pattern as subgraph queries, `--orderBy` and `--orderDirection`?

hopeyen:
Resolves #442 

Updated the actions graphQL schema to include sorting by inputs. Now indexers can supply `orderBy` and/or `orderDirection` when they query for actions. Default: action id, ascending order. 

Example to query through the indexer-cli
```
 ./bin/graph-indexer indexer actions get --type=reallocate --orderBy=amount --orderDirection=desc
```


packages/indexer-common/src/indexer-management/__tests__/actions.ts
```diff
@@ -334,7 +335,7 @@ describe('Actions', () => {
     ).resolves.toHaveProperty('data.actions', [expected])
   })
 
-  test('Queue many actions and retrieve all of a certain status', async () => {
+  test('Queue many actions with certain status and ordering specifications', async () => {
```
fordN, line 338: Not clear from this name what this test is testing.  Makes it seem like it's trying to queue actions in a certain order, but the ordering is when fetching the actions right?

packages/indexer-common/src/indexer-management/__tests__/actions.ts
```diff
@@ -365,11 +366,62 @@ describe('Actions', () => {
...
         })
         .toPromise(),
     ).resolves.toHaveProperty(
       'data.actions',
-      expecteds.sort((a, b) => (a.id > b.id ? -1 : 1)),
+      expecteds.sort((a, b) => (a.id < b.id ? -1 : 1)),
+    )
+  })
+
+  test('Queue many actions with status and invalid ordering specifications', async () => {
```
fordN, line 380: Also not sure what this is testing just from the name. You can queue actions in an invalid order?

packages/indexer-common/src/indexer-management/client.ts
```diff
@@ -167,6 +172,22 @@ const SCHEMA_SDL = gql`
     priority: Int
   }
 
+  input ActionOrderByInput {
```
fordN, line 175: Good clear name, but maybe for consistency with other input names remove `Input` from the end?

packages/indexer-cli/src/commands/indexer/actions/get.ts
```diff
@@ -31,9 +34,11 @@ module.exports = {
 
     const inputSpinner = toolbox.print.spin('Processing inputs')
 
-    const { type, status, source, reason, h, help, o, output } = parameters.options
+    const { type, status, source, reason, orderBy, orderDirection, h, help, o, output } =
+      parameters.options
 
     const [action] = fixParameters(parameters, { h, help }) || []
+    const actionsOrder: ActionOrderByInput = {}
```
fordN, line 41: Don't we want this to be undefined if no `orderBy` argument is provided?

packages/indexer-cli/src/commands/indexer/actions/get.ts
```diff
@@ -75,6 +80,10 @@ module.exports = {
         }
       }
 
+      if (orderBy) {
+        actionsOrder[orderBy as keyof ActionOrderByInput] =
+          orderDirection === Sort.desc ? orderDirection : Sort.asc
```
fordN, line 85: Not totally following the purpose of this line of code. 

Why the `orderDirection === Sort.desc ? orderDirection : Sort.asc`? 

Instead should we just parse the provided `orderDirection` input into a `Sort` enum variant above and set that to the actionsOrder object?

fordN: We can default to `Sort.desc` if no `orderDirection` was provided.

packages/indexer-cli/src/commands/indexer/actions/get.ts
```diff
@@ -90,20 +99,25 @@ module.exports = {
...
+          {
+            type,
+            status,
+            source,
+            reason,
+          },
+          actionsOrder,
+        )
       }
       actionSpinner.succeed('Actions query returned')
```
fordN, line 122: Have you tested out this indexer-cli command on your test indexer?

packages/indexer-cli/src/commands/indexer/actions/get.ts
```diff
@@ -1,7 +1,8 @@
+import { ActionOrderByInput } from './../../../../../indexer-common/src/actions'
```
fordN, line 1: Move this import to be from @graphprotocol/indexer-common? (currently on line 5)

packages/indexer-common/src/indexer-management/__tests__/actions.ts
hopeyen: Yes, right! updating to 'Queue many actions and retrieve all of a certain status with certain ordering', if that's better

packages/indexer-common/src/indexer-management/__tests__/actions.ts
hopeyen: No :( That's a bad name, updating to `Queue many actions and retrieve all of a certain status with invalid ordering`, if that's better. This is meant to test for wrong input for `orderBy`

packages/indexer-common/src/indexer-management/__tests__/actions.ts
fordN: sounds good!

packages/indexer-common/src/indexer-management/__tests__/actions.ts
fordN: sounds good!

packages/indexer-cli/src/commands/indexer/actions/get.ts
hopeyen: cleaned it up a little bit, that was a no-brain kind of parsing the input :/

packages/indexer-common/src/indexer-management/client.ts
```diff
@@ -167,6 +177,39 @@ const SCHEMA_SDL = gql`
...
+    poi
+    force
+    source
+    reason
+    priority
+    createdAt
+    updatedAt
+  }
+
+  input ActionOrderBy {
```
fordN, line 197: Oops, we should remove this now since it's unused.

fordN (approved):
LGTM 👍

Issue #445, opened 2022-07-29
Action queue improvements: Store reason for every action added to queue


fordN:
To improve the experience of managing the queue and more specifically debugging why an item was added to the queue we should include the reason each action was added. i.e - include a matching rule id and the reason why it was deemed worthy of allocating to.  

Something like: `5: decisionBasis=Always` --> meaning it matched on rule id=5 and led to suggest allocating because decisonBasis=Always. 

Another example: `global: minSignal met` --> meaning it fell back to the global rule since no specific rule matched, and the signal threshold was met.

PR #461, opened 2022-08-13
indexer-common,indexer-agent: Include reason for actions being added to queue

fordN:
With the ultimate goal of including meaningful reason strings for each action added to the actions queue a number of changes have been made in this PR. 

The process of resolving which deployments are worth allocating towards has been updated to now result in an array of allocation decisions with length equal to the number of subgraph deployments on the network, including both toAllocate and not toAllocate decisions (rather than an array of only subgraph ids to allocate to).  

```typescript
AllocationDecision {
  declare deployment: SubgraphDeploymentID
  declare toAllocate: boolean
  declare ruleMatch: RuleMatch
  }
```

In addition, the process of resolving allocation actions has been updated. `reconcileAllocations`  and `reconcileDeploymentAllocations` have been replaced with `reconcileActions`,`reconcileDeploymentActions` and in the process been simplified and allowed for concurrency in reconciling many actions.  

A few more structural changes have also been made to move some network querying functions into the `NetworkMonitor`.

### DIFF
--- a/packages/indexer-agent/src/__tests__/indexer.ts
+++ b/packages/indexer-agent/src/__tests__/indexer.ts
@@ -8,6 +8,7 @@ import {
   toAddress,
 } from '@graphprotocol/common-ts'
 import {
+  AllocationManagementMode,
   createIndexerManagementClient,
   defineIndexerManagementModels,
   IndexerManagementClient,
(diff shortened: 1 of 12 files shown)


fordN:
Resolves #445

hopeyen (approved):
LGTM!

PR #463, opened 2022-08-15
indexer-cli: command to batch approve queued

### LINKED ISSUE #449: AQI: Support command to approve all queued actions
Right now the `actions approve` command takes a list of action ids which can be cumbersome to supply especially for indexers with a large number of allocations.  Would be useful to support a command to batch approve all queued actions, like: `graph indexer actions approve queued`

hopeyen:
Resolves #449 

If there's more than one actions with queued status, indexers can use `indexer-cli` as such to approve all the queued actions
```
./bin/graph-indexer indexer actions approve queued
```

If there's no queued actions, expect error `Failed to process action ids, check queue for available actions`.

Alternatively we can add a new query type for `approveActions` to include parameter for `ActionStatus`?

### DIFF
--- a/packages/indexer-cli/src/commands/indexer/actions/approve.ts
+++ b/packages/indexer-cli/src/commands/indexer/actions/approve.ts
@@ -4,10 +4,12 @@ import chalk from 'chalk'
 import { loadValidatedConfig } from '../../../config'
 import { createIndexerManagementClient } from '../../../client'
 import { fixParameters, printObjectOrArray } from '../../../command-helpers'
-import { approveActions } from '../../../actions'
+import { approveActions, fetchActions } from '../../../actions'
+import { ActionStatus } from '@graphprotocol/indexer-common'
 
 const HELP = `
 ${chalk.bold('graph indexer actions approve')} [options] [<actionID1> ...]
+${chalk.bold('graph indexer actions approve')} [options] queued
 
 ${chalk.dim('Options:')}
 
@@ -37,6 +39,8 @@ module.exports = {
 
     let numericActionIDs: number[]
 
+    const config = loadValidatedConfig()
+    const client = await createIndexerManagementClient({ url: config.api })
     try {
       if (!['json', 'yaml', 'table'].includes(outputFormat)) {
         throw Error(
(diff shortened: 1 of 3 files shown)


packages/indexer-cli/src/commands/indexer/actions/approve.ts
```diff
@@ -48,7 +51,18 @@ module.exports = {
...
+      if (actionIDs.join() == 'queued') {
+        const queuedActions = await fetchActions(client, {
+          status: ActionStatus.QUEUED,
+        })
+        numericActionIDs = queuedActions.map(action => action.id)
+      } else {
+        numericActionIDs = actionIDs.map(action => +action)
+      }
+      if (numericActionIDs.length === 0) {
+        throw Error(`Failed to process action ids, check queue for available actions`)
```
fordN, line 64: Not sure users are going to immediately know what went wrong and how to debug from this log message.

What if we move this check into the first if block of the if..else statement above and make it specific to the `queued` actions? (The case of an empty `actionIDs` array is already checked for above.) 

We could then simply log:
```
`No 'queued' actions found.`
```

packages/indexer-cli/src/commands/indexer/actions/approve.ts
```diff
@@ -4,7 +4,8 @@ import chalk from 'chalk'
 import { loadValidatedConfig } from '../../../config'
 import { createIndexerManagementClient } from '../../../client'
 import { fixParameters, printObjectOrArray } from '../../../command-helpers'
-import { approveActions } from '../../../actions'
+import { approveActions, fetchActions } from '../../../actions'
+import { ActionStatus } from '@graphprotocol/indexer-common'
 
 const HELP = `
 ${chalk.bold('graph indexer actions approve')} [options] [<actionID1> ...]
```
fordN, line 11: Can you include the new `queued` variant that can be supplied as an argument to this command?

In other commands we've done this by adding another line to the help with the new command format. Using that paradigm we could add another line like:

```
${chalk.bold('graph indexer actions approve')} [options] queued
```

fordN (approved):
LGTM 👍

PR #465, opened 2022-08-16
Set default for cost models plurals, add tests

hopeyen:
Received report on query to cost model still returns null when global cost model has been set. 

- Update indexer-serivce `costModels` queries to return global cost models for the deployments without cost model defined.
- Add basic tests in `indexer-common` as we debug the potential issue

### DIFF
--- a/packages/indexer-service/src/server/cost.ts
+++ b/packages/indexer-service/src/server/cost.ts
@@ -167,7 +167,7 @@ export const createCostServer = async ({
             const result = await context.client
               .query(
                 gql`
-                  query costModel($deployment: [String!]) {
+                  query costModel($deployment: String!) {
                     costModel(deployment: $deployment) {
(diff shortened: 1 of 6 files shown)


packages/indexer-service/src/server/cost.ts
```diff
@@ -151,7 +151,8 @@ export const createCostServer = async ({
         },
 
         costModel: async (parent, args: CostModelArgs, context: Context) => {
-          const deployment = new SubgraphDeploymentID(args.deployment).bytes32
+          const deploymentID = new SubgraphDeploymentID(args.deployment)
+          const deployment = deploymentID.bytes32
```
fordN, line 155: This appears redundant with line 179? Is it used elsewhere too?

fordN (approved):
👍

PR #470, opened 2022-08-29
Indexer-common: Fix automatic rule creation upon confirmed action execution

fordN:
This PR primarily contains a few important fixes to the auto rule creation that is done after a manual action is executed. 

Fixes included: 
- Use correct number formatting when auto creating rules upon confirmation of action execution (test case added for this).
- Always upsert an `offchain` rule after an `unallocate` action to ensure the agent doesn't recreate the allocation.

Small improvements made while updating:
- Improved detail of error messages when batch execution fails confirmTransactions().
- Add a test for inserting and fetching rules using helper functions in indexer-common (testing the first fix listed above).


packages/indexer-common/src/indexer-management/allocations.ts
```diff
@@ -667,19 +678,17 @@ export class AllocationManager {
...
-        `No matching indexing rule found; updating indexing rules so indexer-agent keeps the deployment synced but doesn't reallocate to it`,
-      )
-      const offchainIndexingRule = {
-        identifier: allocation.subgraphDeployment.id.ipfsHash,
-        identifierType: SubgraphIdentifierType.DEPLOYMENT,
-        decisionBasis: IndexingDecisionBasis.OFFCHAIN,
-      } as Partial<IndexingRuleAttributes>
+    // Upsert a rule so the agent keeps the deployment synced but doesn't allocate to it
+    logger.debug(
+      `Matching indexing rule found; updating indexing rules so indexer-agent keeps the deployment synced but doesn't reallocate to it`,
```
hopeyen, line 683: maybe we don't need to specify `Matching indexing rule found; `?

packages/indexer-common/src/indexer-management/allocations.ts
```diff
@@ -667,19 +678,17 @@ export class AllocationManager {
...
+    )
+    const offchainIndexingRule = {
+      identifier: allocation.subgraphDeployment.id.ipfsHash,
+      identifierType: SubgraphIdentifierType.DEPLOYMENT,
+      decisionBasis: IndexingDecisionBasis.OFFCHAIN,
```
hopeyen, line 688: Should we later on add something to change these deployments to `IndexingDecisionBasis.NEVER`? Perhaps after they are not considered recently closed anymore

packages/indexer-common/src/errors.ts
```diff
@@ -144,12 +144,13 @@ export const INDEXER_ERROR_MESSAGES: Record<IndexerErrorCode, string> = {
   IE060: 'Failed to allocate: Already allocating to the subgraph deployment',
   IE061: 'Failed to allocate: Invalid allocation amount provided.',
   IE062: 'Did not receive tx receipt, not authorized or network paused.',
-  IE063: 'No active allocation with provided id found.',
-  IE064: 'Failed to unallocate: Allocations cannot be closed in this epoch.',
-  IE065: 'Failed to unallocate: Allocation has already been closed.',
-  IE066: 'Failed to allocate: allocation ID already exist on chain',
+  IE063: 'No active allocation with provided id found',
```
hopeyen, line 147: We might as well remove the `.` for IE061 and IE062?

hopeyen (approved):

packages/indexer-common/src/indexer-management/allocations.ts
fordN: Agreed, I'll update.

packages/indexer-common/src/indexer-management/allocations.ts
fordN: I'll add to the todo list, but right now I think it's ok to leave up to the indexer.   Maybe in the meantime we should make note of this in the actions queue guide, so indexers know how to deal with this situation.

packages/indexer-common/src/errors.ts
fordN: Oops, I missed a few .'s.  I'll remove.

PR #471, opened 2022-08-29
indexer-common: allow 0 staked allocations

hopeyen:
Updating logic check in the indexer agent per [GIP0027](https://forum.thegraph.com/t/gip-0027-create-altruistic-allocations-and-close-stale-allocations/3217/6) by simply removing the constraint on 0.


fordN (approved):
LGTM 👍

PR #472, opened 2022-08-29
indexer-common: network pauses exit message

### LINKED ISSUE #207: indexer-agent always exits with fatal error: CALL_EXCEPTION for method: paused()
Versions:
- indexer-agent: 0.10.0
- indexer-service: 0.10.0
- graph-node: 0.21.1

Environment:
- New testnet configuration
- Database is running
- Graph-node is running and scanning blocks from the Ethereum node.
 
Expected behavior: 
After launch container should check all settings and provide notifications like "waiting for the node operator wallet approval from staking contract"  It could be checked in the loop every 1-5 minutes.

Actual behavior:
Service exits with a fatal error. It makes it very hard to verify the configuration.

Logs:
https://gist.github.com/begetan/74d28161bb9ce7eb8fdc15eb1811fc8c

hopeyen:
Early exit message on contract calls to the network

Resolve #207

### DIFF
--- a/packages/indexer-common/src/network.ts
+++ b/packages/indexer-common/src/network.ts
@@ -691,40 +691,48 @@ async function monitorNetworkPauses(
   contracts: NetworkContracts,
   networkSubgraph: NetworkSubgraph,
 ): Promise<Eventual<boolean>> {
-  return timer(60_000)
-    .reduce(async (currentlyPaused) => {
-      try {
-        const result = await networkSubgraph.query(
-          gql`
-            {
-              graphNetworks {
-                isPaused
+  try {
+    return timer(60_000)
+      .reduce(async (currentlyPaused) => {
+        try {
+          const result = await networkSubgraph.query(
+            gql`
+              {
+                graphNetworks {
+                  isPaused
+                }
               }
-            }
-          `,
-        )
+            `,
+          )
 
-        if (result.error) {
-          throw result.error
+          if (result.error) {
+            throw result.error
+          }
+
+          if (!result.data || result.data.length === 0) {
+            throw new Error(`No data returned by network subgraph`)
+          }
+
+          return result.data.graphNetworks[0].isPaused
+        } catch (err) {
+          logger.warn(`Failed to check for network pause, assuming it has not changed`, {
+            err: indexerError(IndexerErrorCode.IE007, err),
+            paused: currentlyPaused,
+          })
+          return currentlyPaused
         }
-
-        if (!result.data || result.data.length === 0) {
-          throw new Error(`No data returned by network subgraph`)
-        }
-
-        return result.data.graphNetworks[0].isPaused
-      } catch (err) {
-        logger.warn(`Failed to check for network pause, assuming it has not changed`, {
-          err: indexerError(IndexerErrorCode.IE007, err),
-          paused: currentlyPaused,
-        })
-        return currentlyPaused
-      }
-    }, await contracts.controller.paused())
-    .map((paused) => {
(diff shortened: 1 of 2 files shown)


packages/indexer-common/src/network.ts
```diff
@@ -607,40 +616,48 @@ async function monitorNetworkPauses(
...
+          })
+          return currentlyPaused
+        }
+      }, await contracts.controller.paused())
+      .map((paused) => {
+        logger.info(paused ? `Network paused` : `Network active`)
+        return paused
+      })
+  } catch (error) {
+    logger.error(`Failed to check for network pause with contracts controller`, {
```
fordN, line 655: Could you explain the double wrapped try..catch here? I would expect the inner one to catch the error and still return valid value (`return currentlyPaused`).  Why is the outer try..catch needed if the inner one doesn't throw an error?

packages/indexer-common/src/network.ts
```diff
@@ -127,31 +127,40 @@ export class Network {
...
+        rebateClaimMaxBatchSize,
+        poiDisputeMonitoring,
+        poiDisputableEpochs,
+        gasIncreaseTimeout,
+        gasIncreaseFactor,
+        baseFeePerGasMax,
+        maxTransactionAttempts,
+      )
+    } catch (error) {
+      throw indexerError(IndexerErrorCode.IE004, error.message)
```
fordN, line 162: Is this try..catch needed since it will still throw an error? Shouldn't the functions here: monitorNetworkPause() and monitorIsOperator() throw useful, specific errors on their own?

hopeyen: this catches the same case on the other comment on `suberror: IndexerErrorCode.IE007` for `monitorNetworkPause()`

Without both catches, we get an error like 
```
Error: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="paused()", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.6.1)
    at Logger.makeError 
``` . 

With the other handle, IE007 is probably enough. I wanted to point people towards IE004 errors markdown because I thought it was helpful but not referenced elsewhere, so the error message would have 
```
[1662231687713] ERROR (IndexerAgent/4838 on book): Failed to check for network pause with contracts controller
    ...
   suberror: "IE007"
    cause: "call revert exception ..."
IndexerError: Failed to synchronize with network
     code: 'IE004',
  explanation: 'https://github.com/graphprotocol/indexer/blob/master/docs/errors.md#ie004',
  cause: 'Failed to check for network pause'
}
```

(I will remove the catch for `IE004` though, I can see how it is redundant

packages/indexer-common/src/network.ts
hopeyen: This is the catch the failure to call  `await contracts.controller.paused()` on [L649](url) when `currentlyPaused` first gets initialized. 
Thinking  definitely be moved to explicitly wrap the contract call. I agree that the inner one doesn't throw an error and easier to follow

tilacog: I left a suggestion for this in #583, which focuses the try/catch scope on the expression that evaluates the network pause.

It is based on this PR.

tilacog (approved):
Tested the indexer-agent locally and it works as expected. 

Similar errors are still around some contract calls, like this call to `stableCoin.decimals()`:

https://github.com/graphprotocol/indexer/blob/8c806154a68f97db8b6614b8d059ab296c0cefe4/packages/indexer-agent/src/cost.ts#L83

But I believe it can be worked in future PRs.

PR #479, opened 2022-09-07
Integrate Epoch Block Oracle

hopeyen:
In preparation for the Epoch Block Oracle (EBO) to be the source of true latest valid epochs and latest valid epoch start blocks in the multi-chain world, This PR integrates the EBO subgraph such that `indexer-agent,common,service` can resolve POIs for subgraph deployments on different networks.

### Usage
- Supply `indexer-agent` startup param endpoint `--block-oracle-endpoint`
- Make sure your graph node has correct ethereum connections (mainnet, goerli, gnosis, etc)

### Updates/Notes
- Hardcoded constant variable of supported network aliasing `NetworkAliases` for Ethereum mainnet, goerli, and gnosis chain. (Remove once aliasing happens on the EBO subgraph or the graph node.  
- When resolving POI for an allocation, query the EBO subgraph for start block number, and then query graph node for the corresponding block hash. 
-- This replaces the previous process: query the network epochManager contract for the current epoch start block, then query the indexer agent's eth client for the block hash.   
- We can replace `currentEpochs` and other `getBlock(...).hash` calls later on


packages/indexer-agent/src/agent.ts
```diff
@@ -441,9 +454,22 @@ class Agent {
...
+        globalLatestValidEpoch,
       }) => {
-        this.logger.info(`Reconcile with the network`)
+        this.logger.info(`Reconcile with the network`, {
+          globalLatestValidEpoch,
+        })
 
+        if (globalLatestValidEpoch.epochNumber != currentEpoch) {
+          this.logger.warn(
+            `EBO latest valid epoch differs from the network contract, ping updates to the EBO (After this is stable, can replace currentEpoch)`,
```
fordN, line 465: 👍

packages/indexer-service/src/commands/start.ts
```diff
@@ -141,6 +142,12 @@ export default {
         required: false,
         group: 'Network Subgraph',
       })
+      .option('block-oracle-endpoint', {
```
fordN, line 145: Does the indexer-service use the block oracle subgraph?

packages/indexer-common/src/errors.ts
```diff
@@ -151,6 +153,8 @@ export const INDEXER_ERROR_MESSAGES: Record<IndexerErrorCode, string> = {
   IE066: 'Failed to allocate: allocation ID already exists on chain',
   IE067: 'Failed to query POI for current epoch start block',
   IE068: 'User-provided POI did not match reference POI from graph-node',
+  IE069: 'Failed to query Epoch Block Oracle',
```
fordN, line 156: This error is specific to the subgraph right? 

`Failed to query Epoch Block Oracle Subgraph`?

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -549,11 +555,113 @@ export class NetworkMonitor {
...
+    }
+  }
+
+  async fetchPOIBlockPointer(allocation: Allocation): Promise<BlockPointer> {
+    try {
+      const networkID = (
+        await this.indexingStatusResolver.indexingStatus([
+          allocation.subgraphDeployment.id,
+        ])
+      )[0].chains[0].network
```
fordN, line 625: Here also, would love to use something more clear than [0] indexing if it makes sense.

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -549,11 +555,113 @@ export class NetworkMonitor {
...
+      const epoch = await this.latestValidEpoch(networkID)
+      const epochStartBlock = {
+        number: epoch.startBlockNumber,
+        hash: epoch.startBlockHash,
+      }
+      // // Previous epoch also acceptable by the arbitration charter, but not sure if we should leave this
+      // // as there could be delay between epoch update and EBO syncs
```
fordN, line 632: I think we'll need to leave in the fallback to previous epoch in case the Epoch Block Oracle has delay in sending the message at start of epoch.

README.md
```diff
@@ -203,6 +203,8 @@ Protocol
                                deployment             [string] [default: "0.01"]
   --register                   Whether to register the indexer on chain
                                                        [boolean] [default: true]
+  --block-oracle-endpoint      Endpoint to query epoch start blocks from
```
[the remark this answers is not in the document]
tilacog: Currently, we refer to it as "the Epoch Subgraph", but I'm not sure if this is official or final.

packages/indexer-agent/src/agent.ts
```diff
@@ -198,6 +199,19 @@ class Agent {
       },
     )
 
+    const globalLatestValidEpoch = timer(600_000).tryMap(
```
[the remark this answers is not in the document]
tilacog: We refer to the network which holds the Graph protocol contracts as the `protocol chain`. 

For every other chain, we call each an `indexed chain`.

packages/indexer-common/src/allocations/types.ts
```diff
@@ -175,3 +176,50 @@ export const parseGraphQLEpochs = (epoch: any): Epoch => ({
...
+  epochNumber: number
+  startBlockNumber: number
+  startBlockHash: string
+  prevStartBlockNumber: number
+  prevStartBlockHash: string
+}
+
+export const NetworkAliases: { [key: string]: string } = {
+  mainnet: 'eip155:1',
+  rinkeby: 'eip155:4',
```
tilacog, line 191: The EBO doesn't currently track rinkeby. Is this a requirement?

hopeyen: No, I will delete this:)

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -549,11 +555,113 @@ export class NetworkMonitor {
...
+      if (!result.data.globalStates) {
+        throw indexerError(
+          IndexerErrorCode.IE069,
+          `Failed to query latest valid epoch number for the global state`,
+        )
+      }
+
+      const epochBlock = await buildEpochBlock(
+        this.indexingStatusResolver,
+        result.data.globalStates[0].latestValidEpoch.epochNumber,
```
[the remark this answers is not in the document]
tilacog: Just a heads up:
We will change the schema a little bit once https://github.com/graphprotocol/graph-node/pull/3096 is merged. The `epochNumber` field will most likely be exclusive to the `Epoch` entity. 

The change won't affect this query, but it might introduce a more convenient way to do it. I can help with this update when that happens.

tilacog (dismissed):
LGTM, considering the Epoch Subgraph scope.
I would say that removing `rinkeby` as an indexed chaing is the only required change on that matter.

packages/indexer-service/src/commands/start.ts
hopeyen: I don't think so. This was added here because I wanted to make the subgraph endpoint a requirement for indexer management server. This requirement can be relaxed or use a default value?

packages/indexer-common/src/indexer-management/monitor.ts
hopeyen: I was thinking that during the EBO's delay period, this fallback would set the latest epoch to the previous of previous epoch and resolve POI from an invalid epoch. So what do you think if I move this fallback to be an error-catch in `resolvePOI` with the current epoch block? we can also only make queries to the previous block numbers when it is needed?

packages/indexer-agent/src/agent.ts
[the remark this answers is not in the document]
hopeyen: Oh that sounds better, I will update to `protocolChainLatestValidEpoch`

packages/indexer-common/src/indexer-management/monitor.ts
hopeyen: There isn't a single element version of the query from the schema as far as I could tell, so I added a query to `indexingStatusForCurrentVersion` with subgraph name

packages/indexer-common/src/indexer-management/monitor.ts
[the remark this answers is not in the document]
hopeyen: Yes, I updated it such that we query only one `globalState` with id specified to 0, so we don't need to use indices

packages/indexer-service/src/commands/start.ts
fordN: I would lean towards making the block oracle subgraph endpoint an optional config.

packages/indexer-common/src/indexer-management/monitor.ts
fordN: I'm sorry but I don't think I'm following exactly.  

If the epoch block oracle subgraph does not yet have data for the current epoch, so effectively the agent cannot know what the start block for current epoch is then it should use the start block from epoch = currentEpoch - 1 to resolve the POI. 

What do you mean by?  
- "set the latest epoch to the previous of previous epoch and resolve POI from an invalid epoch"? 
- "move this fallback to be an error-catch in resolvePOI with the current epoch block"?

README.md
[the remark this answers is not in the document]
fordN: Epoch subgraph works for me!

packages/indexer-service/src/commands/start.ts
fordN: For today I'm ok with setting a default on indexer-service and in subsequent PR we can remove requirement. :)

hopeyen: ahhh just saw the second comment. I have pushed the updates to have optional epoch subgraph endpoint and removed from indexer-service and other unused instances :)

packages/indexer-common/src/indexer-management/monitor.ts
hopeyen: Yes, that makes. Previous my understanding for the delay between new epoch starts and the EBO subgraph update is that the agent would receive old epoch data, and resolve POI with an epoch that's the old previous epoch. Now I understand that EBO subgraph simply won't return data, I also added the extra check with network epoch manager contract

packages/indexer-service/src/commands/start.ts
fordN: Even better! Nice, I'll check it out now.

packages/indexer-common/src/allocations/types.ts
```diff
@@ -175,3 +176,43 @@ export const parseGraphQLEpochs = (epoch: any): Epoch => ({
...
+  startBlockHash: string
+}
+
+export const NetworkAliases: { [key: string]: string } = {
+  mainnet: 'eip155:1',
+  goerli: 'eip155:5',
+  gnosis: 'eip155:100',
+}
+
+export const aliasInverse = (eipFormat: string): string => {
```
fordN, line 193: What does this function actually do? 

It returns an alias right?  What about `networkAlias()`? or `convertToAlias()`? or just `alias()`?

packages/indexer-common/src/allocations/types.ts
```diff
@@ -175,3 +176,43 @@ export const parseGraphQLEpochs = (epoch: any): Epoch => ({
...
+  try {
+    return Object.keys(NetworkAliases).filter(
+      (name) => NetworkAliases[name] == eipFormat,
+    )[0]
+  } catch (error) {
+    // hardcoded alias for now, shouldn't get here
+    // if so this will just let blockHashFromNumber fail later
+    return ''
```
fordN, line 201: Return error? In case a network is not in the `networkAliases` object?

packages/indexer-common/src/allocations/types.ts
hopeyen: Yep, specifically `NetworkAliases` converts the alias (human readable) to the network chain ids, and `aliasInverse` is backwards, converting ids to the alias. 
So I guess, maybe `NetworkAliases` -> `NetworkIDs` and `aliasInverse` -> `convertToAlias`?

fordN (approved):
LGTM 👍

PR #482, opened 2022-09-12
indexer-common: Remove max signal check, towards deprecating max signal

fordN:
The max signal conditional is an unused feature adding unnecessary complexity to the allocation decision logic . This PR removes the usage of it from isDeploymentWorthAllocatingTowards() ahead of deprecating it in the APIs.


hopeyen (approved):
💯

PR #483, opened 2022-09-16
indexer-cli: replace gluegun toolkit print.highlight with chalk colors

### LINKED ISSUE #468: TypeError: print.highlight  is not a function in v20.1 client output
When listing all rules with `graph indexer rules get all` the output is correct however there is an error message at the end of the output:

`TypeError: print.highlight is not a function`

indexer-agent v20.2 (built from docker image)
indexer-cli 20.1(built from stakesquid cli build scripts here https://github.com/StakeSquid/graphprotocol-mainnet-docker/tree/master/cli)

image

hopeyen:
Uncertain dependencies for toolkit's print object being passed around, avoid it by using chalk colors (is it so hard to make life colorful...?:)

resolves #468

### DIFF
--- a/packages/indexer-cli/src/command-helpers.ts
+++ b/packages/indexer-cli/src/command-helpers.ts
@@ -27,6 +27,12 @@
 // parameters array and returns the result of that.
 
 import { table, getBorderCharacters } from 'table'
+
+export enum OutputFormat {
+  Table = 'table',
+  Json = 'json',
+  Yaml = 'yaml',
+}
 import yaml from 'yaml'
 import { GluegunParameters, GluegunPrint } from 'gluegun'
 import { utils } from 'ethers'
@@ -59,11 +65,11 @@ export const fixParameters = (
 export const formatData = (
   // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
   data: any,
-  format: 'json' | 'yaml' | 'table',
+  format: OutputFormat,
 ): string =>
-  format === 'json'
+  format === OutputFormat.Json
     ? JSON.stringify(data, null, 2)
-    : format === 'yaml'
+    : format === OutputFormat.Yaml
     ? yaml.stringify(data).trim()
     : Array.isArray(data)
     ? data.length === 0
@@ -97,13 +103,10 @@ export function pickFields(
   return obj
 }
 
-export function displayObjectData(
-  outputFormat: 'table' | 'json' | 'yaml',
-  data: object,
-): string {
-  return outputFormat === 'json'
+export function displayObjectData(outputFormat: OutputFormat, data: object): string {
+  return outputFormat === OutputFormat.Json
     ? JSON.stringify(data, null, 2)
-    : outputFormat === 'yaml'
+    : outputFormat === OutputFormat.Yaml
     ? yaml.stringify(data).trim()
     : table([Object.keys(data), Object.values(data)], {
         border: getBorderCharacters('norc'),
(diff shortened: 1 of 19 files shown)


packages/indexer-cli/src/command-helpers.ts
```diff
@@ -167,10 +167,13 @@ export async function validatePOI(poi: string | undefined): Promise<string | und
...
-  print: GluegunPrint,
-  outputFormat: 'table' | 'json' | 'yaml',
-): void {
+export function outputFormatting(print: GluegunPrint, outputFormat: string): void {
+  if (!['json', 'yaml', 'table'].includes(outputFormat)) {
+    print.error(`Invalid output format "${outputFormat}"`)
+    process.exitCode = 1
+    return
+  }
+
```
tilacog, line 176: What's the motivation for this change?
Wouldn't it be safer to rely on the type checker for validating the possible values for the `outputFormat` parameter instead of checking it during runtime?

packages/indexer-cli/src/rules.ts
```diff
@@ -212,24 +210,23 @@ export const printIndexingRules = (
...
+    return `${displayIndexingRules(outputFormat, onchainRules)}\n${
+      offchainRules.length
+        ? [
+            chalk.bold('Offchain sync list'),
+            offchainRules.map(rule => {
+              return rule.identifier
+            }),
+          ].join('\n')
+        : chalk.dim(`Not syncing any subgraphs offchain`)
+    }`
```
tilacog, line 222: I'd suggest moving the conditional logic out of this template literal, as it is becoming a bit too convoluted.

tilacog (commented):
Just left a question about the `outputFormat` input validation and requested a minor refactor where we construct the strings to be displayed.

packages/indexer-cli/src/rules.ts
hopeyen: I moved this out to `offchainRulesDisplay` for the Display string with a comment, lmk if that is clear or should be factored out?

packages/indexer-cli/src/command-helpers.ts
tilacog: Sorry, I should clarify that by "safer," I mean type-safe.

The TS compiler would prevent invalid input from being passed to the function in the previous code. I believe this promotes a better developer experience because it removes the need for keeping the validation code inside the function, as it would prevent us from ever sending invalid strings to it. I could summarize this safety claim as "having a compile-time error is preferable to having a runtime error."

Nonetheless, we still need to sanitize the CLI inputs. On this matter, I would suggest that CLI options be validated right after the program starts rather than when the code requires them. This way, we wouldn't worry about invalid values being carried downstream while the program would execute in vain.

packages/indexer-cli/src/rules.ts
tilacog: I liked it! Thanks!
It's easyer to read/review as less stuff is happening in the same statement.

packages/indexer-cli/src/command-helpers.ts
tilacog: Ideally, I'd suggest creating a parser function with a `string -> OutputFormat` signature, where `OutputFormat` is an enum of possible table formats. We would then call this function on every command that accepts the table format input as soon as that value is available to us.

hopeyen: I've updated to a `OutputFormat` enum and parser function for the command inputs, it doesn't utilize the TS compiler with type-safe inputs because I wanted to make it clearly return something undefined for invalid inputs, but let me know if the function should include the possible values as inputs?

tilacog (approved):

PR #484, opened 2022-09-19
Indexer-agent: remove fallback to hardcode default file

hopeyen:
Remove `ensureGlobalCostModel` and `reconcileCostModel` from indexer agent as they clutter the agent without any real changes towards the cost models. 

Now indexers should simply set global default cost model through indexer-CLI such as
```
graph indexer cost set model global [path]/basic.agora
```

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -26,8 +26,6 @@ import {
   RewardsPool,
   Subgraph,
   SubgraphIdentifierType,
-  CostModelAttributes,
-  COST_MODEL_GLOBAL,
   evaluateDeployments,
   AllocationDecision,
 } from '@graphprotocol/indexer-common'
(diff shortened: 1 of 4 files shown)


fordN:
Nice, I like the goal of this PR, but the code changes should be much simpler.

Looking at the code changes though it seems there is other work mixed in with it? It seems to be based off the `https://github.com/graphprotocol/indexer/pull/483` branch?

hopeyen:
> Nice, I like the goal of this PR, but the code changes should be much simpler.
> 
> Looking at the code changes though it seems there is other work mixed in with it? It seems to be based off the `https://github.com/graphprotocol/indexer/pull/483` branch?

You're right! I rebased this branch to main, shouldn't depend on the other PR anymore

fordN (commented):
Looking good!

I see some warning about unused imports (`CostModelAttributes`, and `COST_MODEL_GLOBAL`, `fs`), can we resolve those? Then I think it will be ready to merge.

fordN (approved):
👍

PR #490, opened 2022-10-03
indexer-common: Use consistent deployment names

fordN:
Always create subgraphs with the following naming scheme: `indexer-agent/${deployment.ipfsHash.slice(-10)}`.


hopeyen (approved):
nice👍!

PR #491, opened 2022-10-03
AQI: actions delete command

hopeyen:
- Add `indexer actions delete` command to delete actions by IDs in indexer-cli

example
```
$HELP - graph indexer actions delete [options] [<actionID1> ...]

 ./bin/graph-indexer indexer actions delete 13 16 37
✔ Processed input parameters
✔ 3 actions deleted

 ./bin/graph-indexer indexer actions delete 8 9 10                            
✔ Processed input parameters
✖ [GraphQL] Delete action failed: No action items found with id in [8,9,10]
```
If useful, we can delete all actions by a certain status, but I'm worried if it is too easy? Indexers could also `get` with filters and then feed in action ids as an array


fordN (approved):
Looks good 👍

PR #505, opened 2022-10-11
indexer-common: Trace log data used to evaluate each deployment

fordN:
It's currently a bit tricky to debug allocation decisions. To improve debug process this PR adds an additional trace log at the beginning of `isDeploymentWorthAllocatingTowards()` which includes the deployment's data and the matching rule it is being evaluated against.


hopeyen (approved):
👍

PR #506, opened 2022-10-11
indexer-common: Avoid recreating recently failed actions

tilacog:
Adds a new validation step to `executeQueueOperation`, which prevents the current action from entering the queue if it is found to be a retry from a previously failed attempt.

To do so, we query the database for actions with a `failed` status from the past 24h and check if it contains the current action.

### DIFF
--- a/packages/indexer-common/src/indexer-management/resolvers/actions.ts
+++ b/packages/indexer-common/src/indexer-management/resolvers/actions.ts
@@ -77,18 +77,14 @@ async function executeQueueOperation(
       )
       return updatedAction
     } else {
-      logger.warn(
+      const message =
         `Duplicate action found in queue that effects '${action.deploymentID}' but NOT overwritten because it has a different source and/or status. If you ` +
-          `would like to replace the item currently in the queue please cancel it and then queue the proposed action`,
-        {
-          actionInQueue: duplicateAction,
-          proposedAction: action,
-        },
-      )
-      throw Error(
-        `Duplicate action found in queue that effects '${action.deploymentID}' but NOT overwritten because it has a different source and/or status. If you ` +
-          `would like to replace the item currently in the queue please cancel it and then queue the proposed action`,
-      )
+        `would like to replace the item currently in the queue please cancel it and then queue the proposed action`
+      logger.warn(message, {
+        actionInQueue: duplicateAction,
+        proposedAction: action,
+      })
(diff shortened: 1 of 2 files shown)


packages/indexer-common/src/indexer-management/resolvers/actions.ts
```diff
@@ -24,9 +24,22 @@ async function executeQueueOperation(
   logger: Logger,
   action: ActionInput,
   actionsAwaitingExecution: Action[],
+  recentlyFailedActions: Action[],
   models: IndexerManagementModels,
   transaction: Transaction,
 ): Promise<ActionResult[]> {
+  const recentlyFailedAction = recentlyFailedActions.filter(
+    (a) => a.deploymentID === action.deploymentID && a.type === action.type,
```
fordN, line 32: I think we are going to want to be a bit more strict with checking for a match here. 

If an indexer made some changes the action that is attempting to be added to the queue could be different than the recently failed action and thus not expected to fail. 

It'll be a bit different depending on the action type: 
- ActionType.ALLOCATE - Also check `allocationAmount`
- ActionType.UNALLOCATE - Also check `poi` and `force`
- ActionType.REALLOCATE - Also check `poi`, `force`, and `allocationAmount`

fordN (approved):
LGTM 👍

packages/indexer-common/src/indexer-management/resolvers/actions.ts
```diff
@@ -246,3 +263,28 @@ export default {
     return await actionManager.executeApprovedActions()
   },
 }
+
+// Helper function to assess equality among a enqueued and a proposed actions
+function compareFailedActions(enqueued: Action, proposed: ActionInput): boolean {
+  // actions are not the same if they target different deployments
```
hopeyen, line 269: hey tiago👋  wondering if it would be nice to also check `action.source`, as we only filter if the enqueued action is an action made by `indexerAgent`?

tilacog: Resolved in 53a2cb8.

I've opted to keep this check outside of `compareFailedActions` because we are not strictly comparing for equality.

tilacog:
I changed some variable names in `executeQueueOperation` and added some comments to better clarify that we are performing two different checks.

PR #507, opened 2022-10-12
indexer-common: re-organize error logging

hopeyen:
- Add more explicit error logs
- Avoid querying block information when action (`unallocate` or `reallocate`) is forced and poi is provided.


fordN (approved):
LGTM 👍

PR #510, opened 2022-10-17
Update network parameters for Goerli testnet

tmigone:
Updating network parameters to reflect current values on Goerli testnet:
- Epoch length is 554 blocks which is ~1h 51m.
- Max allocation lifetime is 4 epochs which is ~7h 23m.

These numbers seem odd because they were set pre merge for an average block time of 13 seconds (used to be 2h and 8h).


fordN (approved):
👍

PR #511, opened 2022-10-17
indexer-agent,-common: Use consistent clear network/epoch var naming

fordN:
Main changes: 
- Update network identifier and epoch variable names to improve clarity and consistency as we move to a more complex multi-chain world. 
- Use the user provided network name rather than the JSON-RPC provider returned network name when resolving POIs.

I also included some more general code organization improvements.
- Moved type definitions that have been accumulating in the allocations module in indexer-common to a more general types file. 
- Remove redundant variables in a few places.
- Include components name in logger passed into `ActionManager` and `AllocationManager`.


packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -606,7 +608,7 @@ export class NetworkMonitor {
           }
         `,
         {
-          network,
+          network: networkAlias,
```
hopeyen, line 611: I think the EBO query should receive networkID instead of the alias

hopeyen (commented):
looks good! I think there's just a small change needed

hopeyen (approved):
🙇

PR #512, opened 2022-10-17
indexer-agent,-service: remove index-node configs

### LINKED ISSUE #504: Indexer-agent: Leave index-node assignment decisions to index-nodes
The graph-node.toml config supports flexible deployment rules to direct deployments to specific index-nodes. Specific deployment rules can match on the subgraph name and the network that the deployment is indexing. If matches are not found the deployment assignment are balanced across a list of nodes. 

Let's move logic for balancing deployment assignments across index-nodes out of indexer-agent and allow the graph-node configuration to define the assignment rules. 

Requires that the index-nodes can balance deployments across available index-nodes.  TODO: confirm this functionality before working on this feature.

hopeyen:
Remove indexer agent's assignments for subgraph deployments to specific index nodes, instead encourages the use of graph-node toml configs for flexible deployment rules. 

Currently the indexer agent and service take index-node id as provided at start-up, and round-robin the number of deployments on each node. We would like to make sure that graph node can balance deployments across available index-nodes with flexible deployment rules without conflicts from the agent. 

Resolves #504

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,9 +146,6 @@ Indexer Infrastructure
   --indexer-geo-coordinates             Coordinates describing the Indexer's
                                         location using latitude and longitude
                                    [array] [default: ["31.780715","-41.179504"]]
-  --index-node-ids                      Node IDs of Graph nodes to use for
-                                        indexing (separated by commas)
-                                                              [array] [required]
   --indexer-management-port             Port to serve the indexer management API
                                         at              [number] [default: 8000]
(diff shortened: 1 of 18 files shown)


fordN (commented):
Code looks good!  Let's do some testing in a multi indexing node environment to confirm that deployment assignments get properly balanced across the set of index nodes.

fordN (approved):
👍

PR #513, opened 2022-10-18
indexer-common: Support connecting to any protocol network

fordN:
To support connecting to providers for any network I am updating the handling of the ethereum-network argument on startup. If ethereumNetwork name not provided the provider will use network = 'any'.  This change effectively adds support for non-standard networks with unknown names like arbitrum-goerli, etc....  The user may still specify a network name if they know it: i.e "mainnet" which provides an extra check that the provider is connecting to the correct network. 

While in there I also moved the code to connect to the network provider to a shared indexer-common function for usage by both indexer-agent and indexer-service.


hopeyen (approved):
the refactoring looks great 👍

PR #514, opened 2022-10-18
`indexer-cli`, `indexer-common`: Introduce the `--first` CLI option

tilacog:
This PRs allows the `Action` resolver to limit the number of records returned by the query.

### DIFF
--- a/packages/indexer-cli/src/commands/indexer/actions/get.ts
+++ b/packages/indexer-cli/src/commands/indexer/actions/get.ts
@@ -21,11 +21,12 @@ ${chalk.dim('Options:')}
 
   -h, --help                                                        Show usage information
       --type    allocate|unallocate|reallocate|collect              Filter by type
-      --status  queued|approved|pending|success|failed|canceled     Filter by status 
+      --status  queued|approved|pending|success|failed|canceled     Filter by status
       --source <source>                                             Fetch only actions queued by a specific source
       --reason <reason>                                             Fetch only actions queued for a specific reason
       --orderBy id|deploymentID|amount|priority|...|updatedAt       Order actions by a specific field (default: id)
       --orderDirection asc|desc                                     Order direction (default: desc)
+      --first [N]                                                   Fetch only the N first records (default: all records)
   -o, --output table|json|yaml                                      Choose the output format: table (default), JSON, or YAML
 `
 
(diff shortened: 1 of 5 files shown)


tilacog:
The `actions get` command fails with this message for all possible invocations involving `--first`
```
✖ [GraphQL] Variable "$first" expected value of type "Integer" which cannot be used as an input type.
```

tilacog:
Working as expected now

packages/indexer-cli/src/commands/indexer/actions/get.ts
```diff
@@ -21,11 +21,12 @@ ${chalk.dim('Options:')}
 
   -h, --help                                                        Show usage information
       --type    allocate|unallocate|reallocate|collect              Filter by type
-      --status  queued|approved|pending|success|failed|canceled     Filter by status 
+      --status  queued|approved|pending|success|failed|canceled     Filter by status
       --source <source>                                             Fetch only actions queued by a specific source
       --reason <reason>                                             Fetch only actions queued for a specific reason
       --orderBy id|deploymentID|amount|priority|...|updatedAt       Order actions by a specific field (default: id)
       --orderDirection asc|desc                                     Order direction (default: desc)
+      --first [N]                                                   Fetch only the N first records (defaults to 1, returns all records if omitted)
```
fordN, line 29: What do you think about defaulting to 100? That way we don't cause significant changes in the current default behavior.

packages/indexer-cli/src/commands/indexer/actions/get.ts
```diff
@@ -91,6 +103,22 @@ module.exports = {
           ? OrderDirection[orderDirection.toUpperCase() as keyof typeof OrderDirection]
           : OrderDirection.DESC
       }
+
+      switch (typeof first) {
+        case 'boolean':
+          // can only be `true` at this point
+          fetchFirst = 1
```
fordN, line 110: Similar comment to above. I'd suggest defaulting to `100` to maintain a smooth transition from previous behavior.

fordN (approved):
I'd suggest updating the default behavior. Otherwise looks good!

tilacog:
I'll rebase to tidy up commit history.

fordN (approved):
👍

PR #526, opened 2022-10-26
strict versioning for indexer packages

hopeyen:
To avoid using the package of version 0.25.x that was accidentally published in 368ffd12b6eee4eb7b65aaad9f7357e465df6eba , we define `indexer-common` and `indexer-native` to be strictly using the latest tag `0.20.5-x`


tilacog:
I'll publish the `v0.20.5-alpha.0` version before merging.

PR #528, opened 2022-10-28
docs: update network.md and errors.md

hopeyen:
Update 
- testnet's network parameter and indexer component recommended versions
- Env var for `--epoch-subgraph-endpoint` set for subgraph `graphprotocol/goerli-epoch-block-oracle`
- error.md for IE070

### DIFF
--- a/docs/errors.md
+++ b/docs/errors.md
@@ -872,7 +872,7 @@ Failed to query BlockHashFromNumber from graph node
 
 **Solution**
 
-Graph-node could not find the block hash given network and block number, check if graph-node has access to a network client that has synced to the required block. 
+Graph-node could not find the block hash given network and block number, check if graph-node has access to a network client that has synced to the required block, and ensure indexer agent is connect to the index node configured with an RPC endpoint from the required network
 
 ## IE071
 
--- a/docs/networks.md
+++ b/docs/networks.md
@@ -6,7 +6,7 @@ For mainnet:
 
 | Component       | Release                                                                    |
 | --------------- | -------------------------------------------------------------------------- |
-| contracts       | [1.11.1](https://github.com/graphprotocol/contracts/releases/tag/v1.11.1)    |
+| contracts       | [1.11.1](https://github.com/graphprotocol/contracts/releases/tag/v1.11.1)  |
 | indexer-agent   | [0.18.6](https://github.com/graphprotocol/indexer/releases/tag/v0.18.6)    |
 | indexer-cli     | [0.18.6](https://github.com/graphprotocol/indexer/releases/tag/v0.18.6)    |
 | indexer-service | [0.18.6](https://github.com/graphprotocol/indexer/releases/tag/v0.18.6)    |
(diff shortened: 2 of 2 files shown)


docs/networks.md
```diff
@@ -47,18 +47,19 @@ indexer. The only requirement is a minimum stake of 100k GRT.
...
+| `INDEXER_AGENT_ETHEREUM_NETWORK`            | `--ethereum-network`            | `mainnet`                                                                         |
+| `INDEXER_AGENT_INDEXER_ADDRESS`             | `--indexer-address`             | Ethereum address of mainnet indexer                                               |
+| `INDEXER_AGENT_INDEXER_GEO_COORDINATES`     | `--indexer-geo-coordinates`     | Geo coordinates of mainnet indexer infrastructure                                 |
+| `INDEXER_AGENT_MNEMONIC`                    | `--mnemonic`                    | Ethereum mnemonic for mainnet operator                                            |
+| `INDEXER_AGENT_NETWORK_SUBGRAPH_DEPLOYMENT` | `--network-subgraph-deployment` | `QmV614UpBCpuusv5MsismmPYu4KqLtdeNMKpiNrX56kw6u`                                  |
+| `INDEXER_AGENT_NETWORK_SUBGRAPH_ENDPOINT`   | `--network-subgraph-endpoint`   | `https://gateway.thegraph.com/network`                                            |
+| `INDEXER_AGENT_EPOCH_SUBGRAPH_ENDPOINT`     | `--epoch-subgraph-endpoint`     | `https://api.thegraph.com/subgraphs/name/graphprotocol/goerli-epoch-block-oracle` |
```
fordN, line 59: This should not be set on Mainnet yet, just on Testnet/Goerli, so this line should be moved to ~190.

docs/errors.md
```diff
@@ -872,7 +872,7 @@ Failed to query BlockHashFromNumber from graph node
 
 **Solution**
 
-Graph-node could not find the block hash given network and block number, check if graph-node has access to a network client that has synced to the required block. 
+Graph-node could not find the block hash given network and block number, check if graph-node has access to a network client that has synced to the required block, and that the network is configured in the default index-node.
```
fordN, line 875: What is meant by "default index-node" here?

docs/networks.md
hopeyen: removed! (I imagine this endpoint also works for mainnet in theory?

docs/errors.md
hopeyen: updating this to be `the same index-node`? 
It seems that when indexers query blockHashFromNumber, the network had to be on the same index node, but I guess there's no `default`

fordN: Sorry this still isn't crystal clear to me. Same index node as what?

fordN: Do we mean, ensure that index node is configured with an RPC endpoint from the given network?

docs/networks.md
fordN: Ok, but we need to add it for the testnet configs!

fordN (approved):
👍

PR #547, opened 2022-11-23
indexer-cli: add the --fields option

tilacog:
This PR introduces the `--fields` option to the `indexer actions get` subcommand.

Filed names passed to this option will be used to select the columns displayed in stdout after the Actions query resolves.

For instance, passing the `--fieds id,type,source,transaction` will produce the following output.
```
✔ Processed input parameters
✔ Actions query returned
┌────┬──────────┬──────────────┬────────────────────────────────────────────────────────────────────┐
│ id │ type     │ source       │ transaction                                                        │
├────┼──────────┼──────────────┼────────────────────────────────────────────────────────────────────┤
│ 1  │ allocate │ indexerAgent │ 0x89a4be44e65bde2ef3509f5eeb59e3f33d8fa4b0bb67c015717d41a04051b2a8 │
└────┴──────────┴──────────────┴────────────────────────────────────────────────────────────────────┘
```


fordN (changes requested):
Looks good! If we can remove the need for the hardcoded array of fields that'd be great! 

Also, could you add a few `actions get ...` tests? One for successful use of --fields and at least one for the failure modes? Similar to the rules get tests we already have.

packages/indexer-cli/src/commands/indexer/actions/get.ts
```diff
@@ -27,9 +27,42 @@ ${chalk.dim('Options:')}
       --orderBy id|deploymentID|amount|priority|...|updatedAt       Order actions by a specific field (default: id)
       --orderDirection asc|desc                                     Order direction (default: desc)
       --first [N]                                                   Fetch only the N first records (default: all records)
+      --fields [field1,field2]                                      Comma-separated names of the fields to display
   -o, --output table|json|yaml                                      Choose the output format: table (default), JSON, or YAML
 `
 
+const actionFields: (keyof Action)[] = [
```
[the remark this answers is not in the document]
tilacog: I'm having difficulty figuring out how to access the property names from the `Action` class.

I have tried using `Model.rawAttributes`, [as shown in `sequelize` docs](https://sequelize.org/docs/v6/other-topics/typescript/#getting-a-models-attributes), but all I got was `undefined` instead of a property list.
```ts
export function getAttributeMetadata<M extends Model>(model: ModelStatic<M>): string[] {
  return Object.keys(model.rawAttributes)
}

console.log(getAttributeMetadata(Action))
> undefined
```

I also tried calling `Action.getAttributes`, [as we already do in `indexer-common`](https://github.com/graphprotocol/indexer/blob/main/packages/indexer-common/src/indexer-management/__tests__/resolvers/actions.ts#L144), and this time I got an empty array as a result.
```ts
function extractActionKeys(): string[] {
    const result = []
    for (const actionKey in Action.getAttributes()) {
        result.push(actionKey)
    }
    return result
}
console.log(extractActionKeys())
> []
```

I'm not sure what I'm missing here.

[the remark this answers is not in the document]
tilacog: Would it be OK to perform this validation after we got the query results?
I believe that this should work when dealing with `Action` instances instead of the `Action` class.

tilacog:
As of 14b6176594af5b2c8cce3175e794206f66916e7c, tests fail when interacting with the GraphQL library.

`packages/index-cli/src/__tests__/indexer/actions.test.ts`:

```
- Queueing allocate action, target: QmZZtzZkfzCWMNrajxBf22q7BC9HzoT5iJUK3S8qA6zNZr
✖ [GraphQL] Cannot read properties of undefined (reading 'subgraphDeployment')
```

```
- Querying actions
✖ [GraphQL] Cannot read properties of undefined (reading 'fetchActions')
```

I'm not sure what's the cause of this problem.

This is the query result object, for reference:
```
{
  operation: {
    key: 2586502677,
    query: { kind: 'Document', definitions: [Array], loc: [Location] },
    variables: { filter: {}, orderBy: 'id', orderDirection: 'desc', first: 10000 },
    kind: 'query',
    context: {
      _instance: undefined,
      url: 'http://localhost:18000/',
      fetchOptions: undefined,
      fetch: [Function (anonymous)],
      preferGetMethod: false,
      suspense: false,
      requestPolicy: 'cache-first',
      meta: [Object]
    }
  },
  data: null,
  error: [CombinedError: [GraphQL] Cannot read properties of undefined (reading 'fetchActions')] {
    graphQLErrors: [ [GraphQLError] ],
    networkError: undefined,
    response: Response {
      size: 0,
      timeout: 0,
      [Symbol(Body internals)]: [Object],
      [Symbol(Response internals)]: [Object]
    }
  },
  extensions: undefined,
  hasNext: false
}
```

tilacog:
Included some tests for `--field`, `--first` and a combination of both options.

packages/indexer-cli/src/commands/indexer/actions/get.ts
```diff
@@ -27,9 +27,42 @@ ${chalk.dim('Options:')}
       --orderBy id|deploymentID|amount|priority|...|updatedAt       Order actions by a specific field (default: id)
       --orderDirection asc|desc                                     Order direction (default: desc)
       --first [N]                                                   Fetch only the N first records (default: all records)
+      --fields [field1,field2]                                      Comma-separated names of the fields to display
```
fordN, line 30: I'd add a `,...` in the array to indicate it can be an n length array. Also, since I believe it is required that there are no spaces it would be useful to indicate that as well. 

So it would then be something like: 
```--fields [field1,field2,...] Comma-separated names of the fields to display (no spaces allowed between fields)```

File: packages/indexer-cli/src/commands/indexer/actions/get.ts (hunk 1 above)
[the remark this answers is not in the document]
Comment (fordN) [reply packages/indexer-cli/src/commands/indexer/actions/get.ts:34]: Yeah that sounds good! I'm ok with leaving in the hardcoded array for now though if that would hold this up.

File: packages/indexer-cli/src/commands/indexer/actions/get.ts [hunk 3]
```diff
@@ -107,6 +142,13 @@ module.exports = {
         throw Error(`Invalid value for '--first' option, must have a numeric value.`)
       }
 
+      if (!['undefined', 'string'].includes(typeof fields)) {
+        throw Error(
+          `Invalid value for '--fields' option, must be a comma-separated list of field names.`,
```
fordN, line 147: For formatting consistency let's remove the `.` at the end here.  
(I just found a few other instances of periods used and removed them to maintain this consistent formatting across indexer-cli error messages).

fordN (approved):
Looks great! 
I only have a few message formatting suggestions now. 😄 
...and a reminder to rebase or squash commits before rebase&merge, so we're left with a readable, meaningful commit history!

packages/indexer-cli/src/commands/indexer/actions/get.ts
tilacog: Fixed in 9ea1832.

packages/indexer-cli/src/commands/indexer/actions/get.ts
tilacog: Fixed in bc4e412e3c0eaac221389aec50b4bdee7c42f376.

tilacog:
Squashed and rebased commits for a more concise history

PR #549, opened 2022-11-30
Indexer-cli: add actions delete options

hopeyen:
Quick update to extend the `actions delete` command to allow for deleting `all` actions or by `--status` filter

Ex.
```
graph indexer actions delete all 
graph indexer actions delete --status=queued
```


fordN (approved):
👍

PR #562, opened 2022-12-15
`indexer-common`: Stricter chain indentifier check on `resolveChainId`

tilacog:
Without this extra check, a string containing a `chainId` would be handled as an alias by mistake.


fordN (approved):
👍

PR #576, opened 2023-01-20
Forward graph-attestable header for free queries

Theodus:
Automated fishermen may be relying on free queries to check against untrusted indexer responses. For this purpose, they require the signal that the response is expected to be attestable.


cmwhited (approved):

fordN (approved):
👍

PR #577, opened 2023-01-23
indexer-common: Add avalanche, polygon, celo, & optimism to chain alias lookups

fordN:
Add avalanche, polygon, celo, & optimism to chain alias and ID lookups.

### DIFF
--- a/packages/indexer-common/src/indexer-management/types.ts
+++ b/packages/indexer-common/src/indexer-management/types.ts
@@ -161,6 +161,10 @@ const Caip2ByChainAlias: { [key: string]: string } = {
   hardhat: 'eip155:1337',
   'arbitrum-one': 'eip155:42161',
   'arbitrum-goerli': 'eip155:421613',
+  avalanche: 'eip155:43114',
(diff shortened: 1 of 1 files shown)


packages/indexer-common/src/indexer-management/types.ts
```diff
@@ -161,6 +161,10 @@ const Caip2ByChainAlias: { [key: string]: string } = {
   hardhat: 'eip155:1337',
   'arbitrum-one': 'eip155:42161',
   'arbitrum-goerli': 'eip155:421613',
+  avalanche: 'eip155:43114',
+  polygon: 'eip155:137',
```
tilacog, line 165: Should we also add `matic` for referring to `polygon`?
I'm not sure if that alias is used elsewhere.

tilacog (approved):

packages/indexer-common/src/indexer-management/types.ts
fordN: I think we are going to try to use `polygon` everywhere on the network.

PR #578, opened 2023-01-26
github actions: Add node v17 and v18 support, remove v13

fordN:
The `ci` and `publish-native-binaries` github actions use a matrix of node-versions, running the action for each version of node specified. This PR updates the node version matrix for these github actions, removing support for v13 and adding support for v17 and v18.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,12 +11,16 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-20.04
-
     strategy:
       matrix:
-        node-version: [14, 16]
-
+        node-version: [14, 16, 17]
+        system:
+          - os: ubuntu-20.04
+        include:
+          - node_version: 18
+            system:
+              os: ubuntu-22.04
+    runs-on: ${{ matrix.system.os }}
     services:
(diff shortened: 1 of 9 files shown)


hopeyen (approved):

.github/workflows/ci.yml
```diff
@@ -11,12 +11,16 @@ on:
...
-        node-version: [14, 16]
-
+        node-version: [14, 16, 17]
+        system:
+          - os: ubuntu-20.04
+        include:
+          - node_version: 18
+            system:
+              os: ubuntu-22.04
+    runs-on: ${{ matrix.system.os }}
```
tilacog, line 23: Why Node v15 is not included in the `node-version` matrix?
And why is Node v18 is listed separately?

fordN: I honestly don't remember why it wasn't included, but unless we see specific requests for v15 builds I'm going to just be adding newer versions to the support matrix moving forward. 

The v18 build needs a more recent version of ubuntu; I added it as an `include` so I could have it run on ubuntu-22.04 without changing the `os` of the rest of the build matrix.

PR #591, opened 2023-02-01
Use configured provider for resolving protocol chain blocks

tilacog:
Avoids relying on Graph Node to resolve block hashes for protocol chain blocks.  

This way, Graph Node does not need to be configured with a chain not used for indexing subgraphs.


fordN (approved):
👍

PR #597, opened 2023-02-03
indexer-agent,-common: query fee metrics for Prometheus

hopeyen:
To keep track of query fee metrics, add property `metrics: Metrics` to the receipt collector `AllocationReceiptCollector`, which batch receipts to collect as vouchers (exchanged through partial vouchers if needed), and redeem batched vouchers. 

Added a customized `ReceiptMetrics` property keeps track the count, success rate, and duration for basic query fee collection performances.

```
interface ReceiptMetrics {
	receiptsToCollect: Gauge<'allocation'> 			// Individual receipts to collect
	failedReceipts: Counter<'allocation'> 			// Failed to queue receipts to collect
	partialVouchersToExchange: Gauge<'allocation'> 		// Individual partial vouchers to exchange
	receiptsCollectDuration: Histogram<'allocation'> 	// Duration of processing and exchanging receipts to voucher
	vouchers: Counter<'allocation'> 			// Individual vouchers to redeem
	successVoucherRedeems: Counter<'allocation'> 		// Successfully redeemed vouchers
	invalidVoucherRedeems: Counter<'allocation'> 		// Invalid vouchers redeems - tx paused or unauthorized
	failedVoucherRedeems: Counter<'allocation'> 		// Failed redeems for vouchers
	vouchersRedeemDuration: Histogram<'allocation'> 	// Duration of redeeming vouchers
	vouchersBatchRedeemSize: Gauge<never> 			// Size of redeeming batched vouchers
	vouchersCollectedFees: Gauge<'allocation'> 		// Amount of query fees collected for a voucher'
}
```

Open questions:
- Is it better to track metrics by subgraph deployment id? Tracking by allocation was the easiest but metrics across allocations on the same subgraph deployment could be valuable to track.
- Maybe VouchersCollectedFees gauge can be split into 2, one above the redeem threshold, and one below the redeem threshold.
- ReceiptToCollectDuration currently includes both single shot collection and collection through partial vouchers. Considering that partial vouchers are not used super often I kept them together, but maybe it could become outliers and influence the metrics negatively?

### DIFF
--- a/packages/indexer-common/src/allocations/query-fees.ts
+++ b/packages/indexer-common/src/allocations/query-fees.ts
@@ -1,3 +1,4 @@
+import { Counter, Gauge, Histogram } from 'prom-client'
 import axios from 'axios'
 import {
   Logger,
@@ -6,6 +7,7 @@ import {
   toAddress,
   formatGRT,
   Address,
+  Metrics,
 } from '@graphprotocol/common-ts'
 import {
   Allocation,
@@ -39,8 +41,23 @@ export interface PartialVoucher {
   receipt_id_max: string // (0x-prefixed hex)
 }
 
+interface ReceiptMetrics {
+  receiptsToCollect: Gauge<string>
+  failedReceipts: Counter<string>
+  partialVouchersToExchange: Gauge<string>
+  receiptsCollectDuration: Histogram<string>
+  vouchers: Counter<string>
+  successVoucherRedeems: Counter<string>
+  invalidVoucherRedeems: Counter<string>
+  failedVoucherRedeems: Counter<string>
+  vouchersRedeemDuration: Histogram<string>
+  vouchersBatchRedeemSize: Gauge<never>
+  voucherCollectedFees: Gauge<string>
+}
+
 export interface AllocationReceiptCollectorOptions {
   logger: Logger
+  metrics: Metrics
   transactionManager: TransactionManager
   allocationExchange: Contract
   models: QueryFeeModels
@@ -57,6 +74,7 @@ export interface ReceiptCollector {
 
 export class AllocationReceiptCollector implements ReceiptCollector {
   private logger: Logger
+  private metrics: ReceiptMetrics
   private models: QueryFeeModels
   private transactionManager: TransactionManager
   private allocationExchange: Contract
@@ -70,6 +88,7 @@ export class AllocationReceiptCollector implements ReceiptCollector {
 
   constructor({
     logger,
+    metrics,
     transactionManager,
     models,
     collectEndpoint,
@@ -79,6 +98,7 @@ export class AllocationReceiptCollector implements ReceiptCollector {
     voucherRedemptionMaxBatchSize,
   }: AllocationReceiptCollectorOptions) {
     this.logger = logger.child({ component: 'AllocationReceiptCollector' })
+    this.metrics = registerReceiptMetrics(metrics)
     this.transactionManager = transactionManager
     this.models = models
     this.collectEndpoint = new URL(collectEndpoint)
@@ -166,6 +186,10 @@ export class AllocationReceiptCollector implements ReceiptCollector {
           },
         )
 
+      this.metrics.receiptsToCollect.set(
+        { allocation: receipts[0].allocation },
+        receipts.length,
+      )
       if (receipts.length <= 0) {
         logger.debug(`No receipts to collect for allocation`)
         return false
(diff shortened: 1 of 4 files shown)


tilacog:
I did a quick search for `prom-client` and I found it just in the `yarn.lock` file, not in any `package.json`. Did I miss something or we don't declare it as dependency directly?

hopeyen:
> I did a quick search for `prom-client` and I found it just in the `yarn.lock` file, not in any `package.json`. Did I miss something or we don't declare it as dependency directly?

It should be a dependency from `@graphprotocol/common-ts` which is a dependency of each packages in our repo, do you think it's necessary to declare directly?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -39,8 +41,23 @@ export interface PartialVoucher {
...
+  failedReceipts: Counter<string>
+  partialVouchersToExchange: Gauge<string>
+  receiptsCollectDuration: Histogram<string>
+  vouchers: Counter<string>
+  successVoucherRedeems: Counter<string>
+  invalidVoucherRedeems: Counter<string>
+  failedVoucherRedeems: Counter<string>
+  vouchersRedeemDuration: Histogram<string>
+  vouchersBatchRedeemSize: Gauge<never>
+  voucherCollectedFees: Gauge<string>
```
tilacog, line 55: I don't know much about Prometheus in TS, but why do we need those `string` and `never` as type parameters here?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -166,6 +186,10 @@ export class AllocationReceiptCollector implements ReceiptCollector {
           },
         )
 
+      this.metrics.receiptsToCollect.set(
+        { allocation: receipts[0].allocation },
+        receipts.length,
+      )
```
tilacog, line 192: In this case, why do we use the first receipt allocation?
Is it because we expect them to have been previously filtered with unique allocation ids?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -258,16 +283,17 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
           hint: 'If you would like to redeem vouchers like this, reduce the voucher redemption threshold',
           voucherRedemptionThreshold: formatGRT(this.voucherRedemptionThreshold),
           belowThresholdCount: vouchers.belowThreshold.length,
-          totalValueGRT: formatGRT(
-            vouchers.belowThreshold.reduce(
-              (total, voucher) => total.add(BigNumber.from(voucher.amount)),
-              BigNumber.from(0),
-            ),
-          ),
+          totalValueGRT,
```
tilacog, line 296: Nice refator :+1:

packages/indexer-common/src/allocations/query-fees.ts
tilacog: I assume they are the type signature for the labels, right?

tilacog (approved):
LGTM!
Left a few questions, mostly because I don't know much about prometheus in TS.

tilacog:
I think it would be nice if we add those new metrics in the `CHANGELOG`.

packages/indexer-common/src/allocations/query-fees.ts
hopeyen: Yep, that's my understanding as well, that providing the types here allows us to tag the data with a label of that type later on

packages/indexer-common/src/allocations/query-fees.ts
hopeyen: Yes I think receipts in the same batch are of the same allocation id:
The receipts are filtered with allocation id when grouped into a batch ([line](https://github.com/graphprotocol/indexer/blob/9cf1468afcefc97f91e9e332163cbffdb76ab750/packages/indexer-common/src/allocations/query-fees.ts#L585) of code) and pushed to `receiptsToCollect`, and then the batches are popped off for collecting ([line](https://github.com/graphprotocol/indexer/blob/9cf1468afcefc97f91e9e332163cbffdb76ab750/packages/indexer-common/src/allocations/query-fees.ts#L235))

hopeyen:
> I think it would be nice if we add those new metrics in the `CHANGELOG`.

I was under the impression that we auto-gen changelogs right before releases 💭 I won't mind adding entries if that was the wrong impression. Should I update both indexer-agent and indexer-common?

tilacog:
We use the [´chan´](https://github.com/geut/chan/tree/main/packages/chan) CLI tool to add entries to the CHANGELOG files under each package root, and it wraps all unreleased entries when we do a release.

I think we just need an entry for the new metrics under `indexer-common`.

hopeyen:
> I think we just need an entry for the new metrics under `indexer-common`.

Cool! I added an entry and will merge after test passes

PR #601, opened 2023-02-10
common: split checks and refine error messages when querying the EBO

tilacog:
A very small move towards disambiguating error messages for bad Epoch Subgraph queries.


fordN (approved):
LGTM 👍

PR #602, opened 2023-02-10
indexer-agent: Remove networkCurrentEpoch() from critical path

fordN:
The networkCurrentEpoch() function requires that the protocol chain be registered with the Epoch Block Oracle as an index chain. There may be cases where the protocol chain is not registered as an index chain; we don't want to block the indexer-agent reconciliation loop.


tilacog (approved):
LGTM

PR #604, opened 2023-02-11
common: refactor epoch sync check when resolving current epoch block

tilacog:
Enforces that the Epoch Subgraph is at least one epoch away from the Epoch Manager. 

Also, checks for the catastrophic case when the former is ahead of the latter.

### DIFF
--- a/packages/indexer-common/src/indexer-management/monitor.ts
+++ b/packages/indexer-common/src/indexer-management/monitor.ts
@@ -619,12 +619,47 @@ export class NetworkMonitor {
         )
       }
 
-      // Check for validity with epoch manager currentEpoch, and fetch block hash
-      const validBlock =
-        result.data.network.latestValidBlockNumber.epochNumber ==
-        (await this.currentEpochNumber())
-          ? result.data.network.latestValidBlockNumber
-          : result.data.network.latestValidBlockNumber.previousBlockNumber
+      // Check for validity with Epoch Manager's currentEpoch method, and fetch block hash after that.
+      const epochManagerCurrentEpoch = await this.currentEpochNumber()
+      const epochSubgraphCurrentEpoch =
+        result.data.network.latestValidBlockNumber.epochNumber
+
+      // The Epoch Manager will always lead new epochs, so the epochSubgraphCurrentEpoch should be subtracted.
+      const epochDifference = epochManagerCurrentEpoch - epochSubgraphCurrentEpoch
+
+      const logContext = {
+        epochManagerCurrentEpoch,
+        epochSubgraphCurrentEpoch,
+        epochDifferenceFromContracts: epochDifference,
+        networkID,
+      }
+
+      if (epochDifference < 0) {
+        // Somehow, the Epoch Manager is behind the Epoch Subgraph. This is a critical failure.
+        const criticalErrorMessage = `The Epoch Manager is behind the Epoch Subgraph`
+        this.logger.critical(criticalErrorMessage, { ...logContext })
+        throw new Error(criticalErrorMessage)
+      }
+
+      if (epochDifference > 1) {
+        // It is not acceptable to submit a POI for an epoch more than one epoch away from the current epoch
+        const errorMessage = `The Epoch Subgraph is ${epochDifference} epochs away from the Epoch Manager. \
+It should never be more than 1 epoch away from the Epoch Manager contract. \
+Please submit an issue at https://github.com/graphprotocol/block-oracle/issues/new`
+        this.logger.error(errorMessage, { ...logContext })
+        throw new Error(errorMessage)
+      }
+
+      if (epochDifference === 1) {
+        // It is acceptable that the Epoch Subgraph stays at least one epoch behind the Epoch
+        // Manager, considering this measurement could have happened during an epoch transition.
+        this.logger.info(`Epoch Subgraph is one epoch behind the Epoch Manager`, {
+          ...logContext,
+        })
+      }
+
+      // At this point, the epoch difference is either 0 or 1.
+      const validBlock = result.data.network.latestValidBlockNumber
 
       // Resolve block hash for the given block number.
       // Calls the configured provider for blocks from protocol chain, or Graph Node otherwise.
@@ -639,6 +674,8 @@ export class NetworkMonitor {
       }
 
       const latestBlock = result.data._meta.block.number
+
+      this.logger.info('Resolved current Epoch', { latestBlock, ...logContext })
 
       return {
         networkID,


packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -610,15 +610,43 @@ export class NetworkMonitor {
...
+            epochManagerCurrentEpoch,
+            epochSubgraphCurrentEpoch,
+          })
+          throw new Error(criticalErrorMessage)
+        case epochDifference === 0:
+        case epochDifference === 1:
+          // It is acceptable that the Epoch Subgraph stays at least one epoch behind the Epoch
+          // Manager, considering this measurement could have happened during an epoch transition.
+          validBlock = result.data.network.latestValidBlockNumber
+        case epochDifference === 1:
```
fordN, line 636: This case, `epochDifference === 1`, is mentioned twice.  Is that on purpose?

tilacog: Yes, the intention was to avoid a repeated assignment to `validBlock`, which should happen if `epochDistance` is equal to 0 or 1, but to emit a log only if the distance is equal to 1.

Anyhow, I gave up on `switch` in favor of a `if-else` chain which will be more explicit, although also more verbose.

tilacog:
I ended up with an `if-else` chain because my original idea proved to be hard to control and maintain with a `switch` statement.

I didn't force-pushed to keep means of comparison, but I pretend to squash the merge commit once approved.

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -610,15 +610,43 @@ export class NetworkMonitor {
...
-          : result.data.network.latestValidBlockNumber.previousBlockNumber
-
-      // Resolve block hash for the given block number.
-      // Calls the configured provider for blocks from protocol chain, or Graph Node otherwise.
+      // Check for validity with Epoch Manager's currentEpoch method, and fetch block hash after that.
+      const epochManagerCurrentEpoch = await this.currentEpochNumber()
+      const epochSubgraphCurrentEpoch =
+        result.data.network.latestValidBlockNumber.epochNumber
+
+      // The Epoch Manager will always lead new epochs, so it should be minuend.
```
fordN, line 618: Typo in "minuend", probably meant "minused"? 

How about something like: 
"The Epoch Manager will always lead new epochs, so the epochSubgraphCurrentEpoch should be subtracted"?

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -610,15 +610,43 @@ export class NetworkMonitor {
...
+        this.logger.critical(criticalErrorMessage, {
+          epochManagerCurrentEpoch,
+          epochSubgraphCurrentEpoch,
+        })
+        throw new Error(criticalErrorMessage)
+      }
+
+      if (epochDifference > 1) {
+        // It is not acceptable for the Epoch Subgraph to be more than one epoch away from the
+        // Epoch Manager.
```
fordN, line 633: In this context what's important is meeting the arbitration charter, so to make that clear I'd update the comment slightly: 
"It is not acceptable to submit a POI for an epoch more than one epoch away from the current epoch

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -610,15 +610,43 @@ export class NetworkMonitor {
...
+        const errorMessage = `The Epoch Subgraph is ${epochDifference} epochs away from the Epoch Manager`
```
fordN, line 634: This is a case where we would want indexers to report the issue to us, so we can troubleshoot the epoch block oracle/subgraph. Should we include information on how to report?

Perhaps: 
"The Epoch Subgraph is ${epochDifference} epochs away from the Epoch Manager. It should never be more than 1 epoch away from the Epoch Manager contract. Please submit an issue at https://github.com/graphprotocol/block-oracle/issues/new"

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -610,15 +610,43 @@ export class NetworkMonitor {
...
+
+      // The Epoch Manager will always lead new epochs, so it should be minuend.
+      const epochDifference = epochManagerCurrentEpoch - epochSubgraphCurrentEpoch
+
+      if (epochDifference < 0) {
+        // Somehow, the Epoch Manager is behind the Epoch Subgraph. This is a critical failure.
+        const criticalErrorMessage = `The Epoch Manager is behind the Epoch Subgraph`
+        this.logger.critical(criticalErrorMessage, {
+          epochManagerCurrentEpoch,
+          epochSubgraphCurrentEpoch,
```
fordN, line 626: Include the network alias in all logs in this function?

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -610,15 +610,43 @@ export class NetworkMonitor {
...
+          `Considering the penultimate epoch as the Epoch Subgraph is one epoch behind the Epoch Manager`,
+        )
+      }
+
+      // At this point, the epoch difference is either 0 or 1.
+      const validBlock = result.data.network.latestValidBlockNumber
+
       let startBlockHash: string
       if (networkID == this.networkCAIPID) {
         startBlockHash = (await this.ethereum.getBlock(+validBlock.blockNumber)).hash
```
fordN, line 652: Maybe we can add a log at the end of the function with the results?

```
this.logger.info('Resolved current Epoch', {
    network: networkAlias,
    currentEpoch: validBlock.epochNumber,
    currentEpochstartBlockNumber: +validBlock.blockNumber,
    epochDifferenceFromContracts: epochDifference,
    latestBlock: latestBlock,

File: packages/indexer-common/src/indexer-management/monitor.ts [hunk 7]
```diff
@@ -610,15 +610,43 @@ export class NetworkMonitor {
...
+        const errorMessage = `The Epoch Subgraph is ${epochDifference} epochs away from the Epoch Manager`
+        this.logger.error(errorMessage)
+        throw new Error(errorMessage)
+      }
+
+      if (epochDifference === 1) {
+        // It is acceptable that the Epoch Subgraph stays at least one epoch behind the Epoch
+        // Manager, considering this measurement could have happened during an epoch transition.
+        this.logger.info(
+          `Considering the penultimate epoch as the Epoch Subgraph is one epoch behind the Epoch Manager`,
```
fordN, line 643: Maybe we can simplify this considering my suggested additional log below?

"Epoch Subgraph is one epoch behind the Epoch Manager"

fordN (changes requested):
Code looks good! I made some suggestions for the logging.

packages/indexer-common/src/indexer-management/monitor.ts
tilacog: Minuend is the intended word.
(minuend−subtrahend=difference)
But I can rephrase it, no worries.

tilacog: Fixed in 65380140

packages/indexer-common/src/indexer-management/monitor.ts
tilacog: Fixed in 65380140

packages/indexer-common/src/indexer-management/monitor.ts
tilacog: Fixed in 1b6b1403

packages/indexer-common/src/indexer-management/monitor.ts
tilacog: Fixed in 1b6b1403

packages/indexer-common/src/indexer-management/monitor.ts
tilacog: Fixed in 299e359c

fordN (approved):
👍

tilacog:
Force pushed to squash irrelevant commits

PR #618, opened 2023-03-09
indexer-service: support multiple client signer addresses

pcarranzav:
A future improvement would be to know, from the message, who the expected signer is, rather than looping through the possible values.


pcarranzav:
CI seems to be failing the same as in #615 - seems that an RPC provider is throttling us, which causes the tests to time out?

PR #623, opened 2023-03-14
common: account rewards when resolving unallocate action delta

### LINKED ISSUE #620: Feasibility check should consider available Rewards
The current stake feasibility check should also consider the amount of available rewards in the balance.

tilacog:
Fixes #620.

### DIFF
--- a/packages/indexer-common/src/indexer-management/allocations.ts
+++ b/packages/indexer-common/src/indexer-management/allocations.ts
@@ -75,6 +75,7 @@ export interface ResolvedAction {
   action: Action
   allocates: BigNumber
   unallocates: BigNumber
+  rewards: BigNumber
   balance: BigNumber
 }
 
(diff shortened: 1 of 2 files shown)


tilacog:
I suspect we might be being rate-limited, and this could be causing some tests to fail.

packages/indexer-common/src/indexer-management/allocations.ts
```diff
@@ -1131,27 +1133,36 @@ export class AllocationManager {
...
       }
+
       // Fetch the allocation on chain to inspect its amount
       const allocation = await this.networkMonitor.allocation(action.allocationID)
 
-      /* We intentionally don't check if the allocation is active now because it will be checked
-       * later when we prepare the transaction */
+      // Accrue rewards, except for null or zeroed POI
+      rewards =
+        !action.poi || action.poi === '0' || action.poi === '0x0'
```
fordN, line 1153: When comparing against a "zero" POI I'd suggest using `hexlify()` to generate the comparison value. 

Like: `action.poi === utils.hexlify(Array(32).fill(0))`

fordN (changes requested):

packages/indexer-common/src/indexer-management/allocations.ts
tilacog: Fixed in c8b6254

fordN (approved):
👍

tilacog:
Force pushed for rebasing

PR #628, opened 2023-03-16
Validate Network Subgraph `chainId`

### LINKED ISSUE #555: Network Subgraphs robustness check
It would be ideal if the Indexer Agent checked if all network subgraphs (Epoch Subgraph, Network Subgraph) are based on the same network as it's Ethereum provider is set. 

To do so, we would need:
1. to query the subgraph's for their deployment hash: `query { _meta { deployment } }`
2. an IPFS client to fetch those subgraph's manifests and scrape the `network` field.

tilacog:
Resolves #555 

TODO:
- [x] validate `chainId` when the indexer syncs the Network Subgraph
- [ ] validate `chainId` when the indexer does not syncs the Network Subgraph

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -491,18 +491,22 @@ export default {
       statusEndpoint: argv.graphNodeStatusEndpoint,
     })
 
+    // Parse the Network Subgraph optional argument
+    const networkSubgraphDeploymentId = argv.networkSubgraphDeployment
+      ? new SubgraphDeploymentID(argv.networkSubgraphDeployment)
+      : undefined
+
     const networkSubgraph = await NetworkSubgraph.create({
       logger,
       endpoint: argv.networkSubgraphEndpoint,
-      deployment: argv.networkSubgraphDeployment
-        ? {
-            indexingStatusResolver: indexingStatusResolver,
-            deployment: new SubgraphDeploymentID(
-              argv.networkSubgraphDeployment,
-            ),
-            graphNodeQueryEndpoint: argv.graphNodeQueryEndpoint,
-          }
-        : undefined,
+      deployment:
+        networkSubgraphDeploymentId !== undefined
+          ? {
+              indexingStatusResolver: indexingStatusResolver,
+              deployment: networkSubgraphDeploymentId,
+              graphNodeQueryEndpoint: argv.graphNodeQueryEndpoint,
+            }
+          : undefined,
     })
 
     const networkProvider = await Network.provider(
@@ -708,9 +712,7 @@ export default {
       indexerAddress,
       allocationManagementMode,
     )
-    const networkSubgraphDeploymentId = argv.networkSubgraphDeployment
-      ? new SubgraphDeploymentID(argv.networkSubgraphDeployment)
-      : undefined
+
     if (networkSubgraphDeploymentId !== undefined) {
       // Make sure the network subgraph is being indexed
       //
@@ -720,20 +722,20 @@ export default {
         `graphprotocol/network-subgraph/${networkSubgraphDeploymentId.ipfsHash}`,
         networkSubgraphDeploymentId,
       )
-    }
 
-    // Validate if the Network Subgraph belongs to the current provider's network.
-    // This check must be performed after we ensure the Network Subgraph is being indexed.
-    try {
-      await validateNetworkId(
-        networkMeta,
-        argv.networkSubgraphDeployment,
-        indexingStatusResolver,
-        logger,
-      )
-    } catch (e) {
-      logger.critical("Failed to validate Network Subgraph. Exiting.", e)
-      process.exit(1)
+      // Validate if the Network Subgraph belongs to the current provider's network.
+      // This check must be performed after we ensure the Network Subgraph is being indexed.
+      try {
+        await validateNetworkId(
+          networkMeta,
+          argv.networkSubgraphDeployment,
+          indexingStatusResolver,
+          logger,
+        )
+      } catch (e) {
+        logger.critical('Failed to validate Network Subgraph. Exiting.', e)
+        process.exit(1)
+      }
     }
 
     // Monitor ETH balance of the operator and write the latest value to a metric
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -556,7 +556,7 @@ export default {
     const epochSubgraph = await EpochSubgraph.create(argv.epochSubgraphEndpoint)
 
     const networkMonitor = new NetworkMonitor(
-      await resolveChainId(networkMeta.chainId),
+      resolveChainId(networkMeta.chainId),
       contracts,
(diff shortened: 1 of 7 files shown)


packages/indexer-common/src/indexing-status.ts
```diff
@@ -16,6 +16,14 @@ export interface SubgraphDeploymentAssignment {
   node: string
 }
 
+export interface SubgraphFeatures {
+  // `null` is only expected when there are errors in the Subgraph Manifest.
+  network: string | null
+  // manifest validation errors
+  errors: string[]
+  features: 'nonFatalErrors' | 'grafting' | 'fullTextSearch' | 'ipfsOnEthereumContracts'
```
fordN, line 24: It doesn't look like the `errors` and `features` properties are ever used?

packages/indexer-common/src/indexer-management/__tests__/helpers.ts
```diff
@@ -243,27 +243,27 @@ describe('Actions', () => {
   })
 })
 describe('Types', () => {
-  test('Fail to resolve chain id', async () => {
-    await expect(resolveChainId('arbitrum')).rejects.toThrow(
-      'Failed to resolve CAIP2 ID from the provided network alias: arbitrum',
+  test('Fail to resolve chain id', () => {
+    expect(() => resolveChainId('arbitrum')).toThrow(
+      /^Failed to resolve CAIP2 ID from the provided network alias: arbitrum$/,
```
fordN, line 248: Could you explain the change here from string (wrapped in single quotes) to the new string without quotes and regex looking prefix/suffix?

fordN (commented):
Looking good overall! Just left a few questions around types and tests in my review.

packages/indexer-common/src/indexing-status.ts
tilacog: Positive.
We don't use those fields in the rest of the code.
Removed them in e1b4e9a1.

packages/indexer-common/src/indexer-management/__tests__/helpers.ts
tilacog: Fixed in a0848612

packages/indexer-common/src/indexer-management/__tests__/helpers.ts
```diff
@@ -243,27 +243,27 @@ describe('Actions', () => {
...
+  test('Resolve chain alias: `eip155:1`', () => {
+    expect(resolveChainAlias('eip155:1')).toBe('mainnet')
   })
 
-  test('Fail to Resolve chain alias: `eip155:666`', async () => {
-    await expect(resolveChainAlias('eip155:666')).rejects.toThrow(
-      "Failed to match chain id, 'eip155:666', to a network alias",
+  test('Fail to Resolve chain alias: `eip155:666`', () => {
+    expect(() => resolveChainAlias('eip155:666')).toThrow(
+      /^Failed to match chain id, 'eip155:666', to a network alias in Caip2ByChainAlias$/,
```
hopeyen, line 266: similar comment as above on `string -> regex`, if you want to switch back?

packages/indexer-agent/src/commands/start.ts
```diff
@@ -721,6 +722,19 @@ export default {
       )
     }
 
+    // Validate if the Network Subgraph belongs to the current provider's network.
+    // This check must be performed after we ensure the Network Subgraph is being indexed.
+    try {
+      await validateNetworkId(
+        networkMeta,
+        argv.networkSubgraphDeployment,
```
hopeyen, line 730: would it be nice to check `argv.networkSubgraphEndpoint` for its deployment hash if indexer isn't syncing the network subgraph? Maybe we could expect everyone to be indexing it once there's no hosted service endpoint?

packages/indexer-common/src/indexer-management/__tests__/helpers.ts
tilacog: Thanks, fixed in 341e8d43

packages/indexer-agent/src/commands/start.ts
tilacog: I think it would be an excellent check! 

I thought we always enforced that the Network Subgraph should be synced, but now I see that `argv.networkSubgraphDeployment` can be null.
I'll find out where we could insert that check.

tilacog: Upon reflection, testing the hash of the Network Subgraph is not feasible, as it will only be not indexed if the indexer does not provide its hash. 

Unless I'm missing something, we would not have a reference value to check even if we attempted to query the deployment hash from the endpoint.

tilacog: I think this surfaced the fact that we only run `validateNetworkId` if the indexer syncs the Network Subgraph. 
We should also validate when it does not.

hopeyen: If we want to graph the valid hash of the Network Subgraph when the indexer doesn't provide deploymentHash nor sync the network subgraph but did provide an endpoint, then Network Subgraph would make a request (`query { _meta { deployment } }`) to the endpointClient. I think the decision is made [here](https://github.com/graphprotocol/indexer/blob/1da01fc5b268aa8bd54c604f6d85befc4e39c59d/packages/indexer-common/src/network-subgraph.ts#L149). Do you think this would help?

tilacog: Thanks!
This looks like the perfect spot for the check. But we don't have an IPFS interface to retrieve the subgraph manifest and verify the network name.

Would adding IPFS as a dependency just for this check be too burdensome for indexers?
What do you think we should do in this situation, @fordN?

packages/indexer-common/src/errors.ts
```diff
@@ -159,6 +160,7 @@ export const INDEXER_ERROR_MESSAGES: Record<IndexerErrorCode, string> = {
   IE070: 'Failed to query latest valid epoch and block hash',
   IE071: 'Add Epoch subgraph support for non-protocol chains',
   IE072: 'Failed to execute batch tx (contract: staking)',
+  IE073: 'Failed to query subgraph features',
```
fordN, line 163: We could probably provide a bit more detail here to help with debugging. Many indexers may not be familiar with the subgraph features query, so we could also include a reference to the indexing status endpoint on graph-node so they know more context.  

What about: 'Failed to query subgraph features from indexing statuses endpoint'?

packages/indexer-agent/src/commands/start.ts
fordN: I agree with your proposed approach, Tiago.  Let's work on validating external subgraph endpoints as a separate task.

hopeyen (approved):
lgtm! Just have one small consideration on if this feature should be tested as part of a testnet release first, but I think it does look good

tilacog:
Since it's a startup check, I can test it locally:
- [x] valid setup
- [ ] invalid: different networks for JSON RPC provider and Network Subgraph
- [ ] no-op: no Network Subgrpah deployment hash provided

PR #635, opened 2023-03-31
Make subgraph rename migration failsafe and limit concurrency

tilacog:
Graph-Node seems to reject redeployments when there's a graft involved (reason still unclear).
This causes migration 11 fail and prevents the Agent from starting up.

This PR allows for failed redeployments and emits a warning message to inform the user of the failed redeployment. 
It should enable the Agent to start up successfully despite redeployment failures.

Note for reviewers:
- pMap needed to be upgraded to v5.5.0 so I could use `pMapSkip`.


tilacog:
Closed in favor of #636.

PR #636, opened 2023-03-31
Remove subgraph renaming migration

tilacog:
Errors were encountered when redeploying grafted subgraphs.
We are dropping this migration so the Agent can start successfully.


fordN (approved):
👍

PR #637, opened 2023-03-31
Flexible subgraph deployment name resolution

tilacog:
PR #615 inadvertently caused the indexer Agent to only handle subgraphs registered in the Network Subgraph, as it can no longer name a subgraph deployment if it's not registered there.

This PR reintroduces a simpler naming scheme based only on the IPFS hash,  similar to the previous naming convention, to handle the case of such subgraphs.

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -27,7 +27,7 @@ import {
   SubgraphIdentifierType,
   evaluateDeployments,
   AllocationDecision,
-  formatDeploymentName,
+  resolveSubgraphDeploymentName,
 } from '@graphprotocol/indexer-common'
 import { Indexer } from './indexer'
 import { AgentConfig } from './types'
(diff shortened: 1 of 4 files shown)


fordN:
Nice! I remember there was also an issue with grafted subgraphs being renamed.  What's the plan to make those work?

tilacog:
From what we've seen, the problem with grafting seems to be tied to Graph Node's redeployments. So it's likely that the fix will come from that end.

The error some indexers were seeing was:
```
"message": "subgraph validation error: [the graft base is invalid: deployment not found: Qm....]"
```

It looks like Graph Node validates before it realizes it already has the deployment.

The error shouldn't occur again as the [migration that caused the problem has been discarded](https://github.com/graphprotocol/indexer/pull/636), and this PR won't reintroduce that behavior.

tilacog:
Related: https://github.com/graphprotocol/graph-node/issues/4534

fordN (approved):
👍

fordN:
Looks good! Let's hold off on merging until we have a migration that will rename and delete the old names?

tilacog:
I think we can close this pull request because #644 has undone most of the work that it was based on. 

However, once the issue with Graph-Node's redeployment validation is fixed (https://github.com/graphprotocol/graph-node/issues/4534), we can bring back the branches/PRs that were previously reverted.

PR #642, opened 2023-04-12
indexer-service: Update index-node GraphQL server proxy schema

fordN:
Update schema filter to include root fields needed by the GraphIX data comparison and debugging tool.


neysofu (approved):
Nice! 🙌🏻

PR #644, opened 2023-04-12
Partially reverts PR #615

tilacog:
Partially reverts PR #615, which prominently introduced a new naming scheme for subgraph deployments. 

Due to https://github.com/graphprotocol/graph-node/issues/4534, the experience of migrating existing subgraph names was sub-optimal, so we are deferring the new deployment naming scheme to when we can cleanly perform a complete migration.


fordN (approved):
👍

Issue #652, opened 2023-04-21
Refactor the Network component to hold two different networks


tilacog:
Depends on #651.

Refactor code to create a `Networks` wrapper type that groups `Network` components by chain id.

This allows selecting the appropriate `Network` value based on the `protocolNetwork` property for every context.

E.g., _Action Workers_ must resolve their Action's protocol network before executing them.

tilacog:
Resolved by #668

PR #659, opened 2023-05-12
Add a `protocolNetwork` field to Rules, Actions, POI Disputes and Allocations

### LINKED ISSUE #650: Refactor configuration
Adapt all startup parameters to support multiple protocol networks.

Options changes:

- **ethereum** —> **network-providers**
    - format: list of `network_identifier`:`network_provider_url`
    - still support the old argument name, `ethereum`, mark it as deprecated and move towards removing it
- **network-subgraph-urls**
    - format: list of `network_identifier`:`network_subgraph_url`
- **epoch-subgraph-urls**
    - format: list of `network_identifier`:`epoch_subgraph_url`
- note: `network_identifier` can support both numeric CAIP2 ids OR their string aliases
- Validations:
    - lists provided to the three protocol network options must have same length
    - user must provide all three above values for each `network_identifier`
    - if no `network_identifier` provided, fall back to older behavior, can only use one network in that case.

### LINKED ISSUE #651: Add a `protocolNetwork` column to Rules and Actions table
Depends on #650.

1. Create a migration with these steps:
    1. Alter the tables: `indexingRules` and `actions:` 
        1. add `protocol_network` column, which defaults to null
    2. Resolve every row protocol_network (There can be no `nulls` left behind!)
        - This step can be difficult because we need to consider edge cases, such as when an indexer has allocations in the same subgraph for both layers.
        - derive protocol_network for every rule and action (only need to do for status = queued or approved)
            - gather rules, active allocations (across protocol networks), and actions: matching rules to active allocations to identify the network
    3. Alter the tables again to add a `NOT NULL`

2. Add the `protocolNetwork` field to Rules and Actions ORM models.
    - Also, add it to non-ORM types, like `SubgraphDeployment`.
    - At this point, the compiler will surface every usage of the new field, we might make some wild discoveries.

3. Update the indexer-cli sub-commands to require a protocol network identifier.
    1. **indexingRules**
        1. add the option to specify the protocol network (optional). If not specified, the value will resolve to the default rule
        2. example: `indexer rules set <DeploymentId> protocolNetwork homestead allocationAmount 10000`
    2. **actions**
        1. update actions `queue` commands to require protocol network
        2. example: `indexer actions queue allocate QmYN4ofRb5CUg1WdpLhhNTVCuiiAt29hBKGjTnnxYh9zYt homestead 1000`
    3. **allocations**
        1. update `create` command 
            1. `indexer allocations create <deployment-id> <protocol-network> <amount> <index-node>`
            2. NOTE: validate that allocation ids are unique per-protocol network; if not, we may need to require protocol-network for `allocations close` and `allocations reallocate` and `allocations get`
        2. Update `get` command
            1. support filtering by `protocol_network`
    4. **disputes**
        1. Update `get` command
            1. Support filtering by `protocol_network`
    5. ********cost********
        1. OPEN QUESTION: do we continue to use just deployment_id as the primary key, or do we also update to include the network?
            1. To state that another way: do we need cost models for each deployment to be different per network?
            2. If we do: we’ll need to update the `set`, `get`, and `delete` commands also to require `protocol_network` arg

tilacog:
Built on top of #654.
Resolves #650 , resolves #651

> This PR refactors changes from #654, most notably the input validation.
> Reviewing this PR instead of #654 for the latest refactorings is advised.

Adds a `protocolNetwork`field to multiple types across packages:
- `IndexingRule`
- `POIDispute`
- `Action`
- `Allocation`

Updates cover:
- **database**: a new migration:
  - introduces the new `protocolNetwork`,
  - populates it with the value of the configured network identifier column, and
  - alters primary keys of the `IndexingRules` and `POIDispute` tables
- **Indexer Management GraphQL API** has new identifier types to accommodate the composite primary keys.
- **CLI**, which now requires a mandatory `--network` option for most of its subcommand.

**Changes not introduced in this PR:**
Although affected components now have a `protocolNetwork` value, the program has no awareness of network multiplicity. #652 will address this issue.


tilacog:
Closing in favor of #668

PR #661, opened 2023-05-15
Include `action` and `allocation` IDs on logs for IE053

tilacog:
This PR adds a bit more context around [IE053](https://github.com/graphprotocol/indexer/blob/main/docs/errors.md#ie053) by including the action and the allocation IDs in the log message, so users can determine which action has failed.

The need for this PR arose from indexers needing clarification due to an action appearing to work despite this error being logged. Putting the `actionID` in the error context will ease investigations about such behavior, particularly in distinguishing between successful and failed actions, possibly indicating a retry.


hopeyen (approved):
💯

PR #660, opened 2023-05-15
update networks docs

trader-payne:
**What changed:**
- renamed mainnet.md -> ethereum-mainnet.md
- renamed testnet.md -> ethereum-goerli.md
- bumped indexer stack versions
- use hosted service urls for the network subgraphs

**What's been added:**
- added arbitrum-one.md
- added arbitrum-goerli.md

**What's missing:**
- arbitrum one and goerli network subgraph IPFS hash
- arbitrum one and goerli Data Edge and Block Oracle Owner contract addresses
- arbitrum one and goerli DAI contract address

PR #664, opened 2023-05-17
Release Graph Node `v0.31.0` on mainnet

neysofu:
Still a draft PR because we're waiting for GGP approval.

EDIT 2023-06-23: now ready for review.

### DIFF
--- a/docs/feature-support-matrix.md
+++ b/docs/feature-support-matrix.md
@@ -27,7 +27,7 @@ An example:
 | ENS                      |         | Yes         | Yes          | No                | No                   | No               |
 | File data sources: IPFS  |         | Yes         | Yes          | No                | Yes                  | Yes              |
 
-The accepted `graph-node` version range is also specificied, with an "upgrade window" from the previous version.
+The accepted `graph-node` version range is also specificied, with an "upgrade window" from the previous version. Here's an example:
 
 ```
 graph-node: ≥0.30.0 <0.31.0
--- a/docs/feature-support-matrix.md
+++ b/docs/feature-support-matrix.md
@@ -27,14 +27,11 @@ An example:
 | ENS                      |         | Yes         | Yes          | No                | No                   | No               |
 | File data sources: IPFS  |         | Yes         | Yes          | No                | Yes                  | Yes              |
 
-The accepted `graph-node` version range is also specificied, with an "upgrade window" from the previous version. Here's an example:
+The accepted `graph-node` version range is also specificied; it always comprises of the latest available version and the one immediately preceding it. Here's an example of a version range which includes both 0.30.X and 0.31.X:
 
 ```
-graph-node: ≥0.30.0 <0.31.0
-valid from: 787
-upgrade window: 795
+graph-node: >=0.30 <0.32
 ```
-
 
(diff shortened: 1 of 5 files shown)


PedroMD:
Hi @neysofu - we've recently discussed having one single feature matrix, as we currently have it in 2 files: 

- [feature-support-matrix.md](https://github.com/graphprotocol/indexer/blob/main/docs/feature-support-matrix.md) (outdated)
- [mainnet.md](https://github.com/graphprotocol/indexer/blob/main/docs/networks/mainnet.md) (canonical, ratified by the Council)

We should remove the one from `mainnet.md` and use only the dedicated one (`feature-support-matrix.md`). There's also currently no need for having separate matrixes for Ethereum mainet or Arbitrum One environment, which simplifies things (we may have different .md files for each environment, as proposed here https://github.com/graphprotocol/indexer/pull/660). 

Could you update this PR, please?

PedroMD:
Only because mainnet should be broken down into two files (Ethereum and
Arbitrum One) as each environment requires different configuration
parameters for Indexers. Once that happens, there shouldn't be a need for
separate matrixes? The matrix is the same for both environments (at least
for now). Just a proposal, really.

neysofu:
Sounds good! I've updated this PR and marked it ready for review. I think it's best to wait until https://github.com/graphprotocol/indexer/pull/660 is merged before removing any of the two `.md` files we have for now, as I don't want to introduce conflicts, but I'm all for removing it. Will follow-up once https://github.com/graphprotocol/indexer/pull/660 is merged.

azf20:
@PedroMD @neysofu [the GGP was approved](https://snapshot.org/#/council.graphprotocol.eth/proposal/0x80c55bb8697d16fedb71ccdce40704f24e931cc28f289a029e0717f3b729e6a8), seems that we have some conflicts to resolve?

docs/feature-support-matrix.md
```diff
@@ -27,7 +27,7 @@ An example:
 | ENS                      |         | Yes         | Yes          | No                | No                   | No               |
 | File data sources: IPFS  |         | Yes         | Yes          | No                | Yes                  | Yes              |
 
-The accepted `graph-node` version range is also specificied, with an "upgrade window" from the previous version.
+The accepted `graph-node` version range is also specificied, with an "upgrade window" from the previous version. Here's an example:
```
neysofu, line 30: I'm making it more clear that the `graph-node` version range here is just an example and not prescriptive, i.e. it shouldn't be updated for every new version.

PedroMD:
Do we have an agreement we can waive the upgrade window to make things easier, at least for now? It has always been marked as optional in the ratified [GIP 008](https://snapshot.org/#/council.graphprotocol.eth/proposal/0xbdd884654a393620a7e8665b4289201b7542c3ee62becfad133e951b0c408444) where this feature matrix was initially proposed.

I also just noticed the GGP the Council signed needed to be updated; it does not reflect the breakdown I had proposed for mainnet and Arbitrum mainnet. Governance-wise, we should proceed as-is anyway and ensure that next time we have the breakdown for each environment.

azf20 (approved):

PR #676, opened 2023-07-18
Updated feature support matrix to support indexing rewards for substreams 

PedroMD:
I also removed the feature support matrix from Ethereum mainnet & Arbitrum One environments, as we've agreed it should be the same on environments.

Another GGP will be created in https://snapshot.org/#/council.graphprotocol.eth referencing the latest feature support matrix. After Council approval, this PR should be merged.


PedroMD:
@azf20 the related GGP-0025 has passed. I did change the .md a bit, not just the matrix itself. Lmk if you think this can be merged already.

azf20 (changes requested):
Nice - I think this just needs an update to include a link to this matrix here https://github.com/graphprotocol/indexer/blob/main/docs/networks.md

azf20 (approved):

PR #677, opened 2023-07-19
agent,common: Automatic support of deployment transfers to L2

fordN:


packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -489,6 +490,50 @@ export class NetworkMonitor {
...
+
+        // There may be no transferred subgraphs, handle gracefully
+        if (transferredDeployments.length == 0) {
+          this.logger.warn('Failed to query subgraph deployments transferred to L2: no deployments found')
+          throw new Error('No transferred subgraph deployments returned')
+        }
+
+        return transferredDeployments.map((deployment: any) => {
+          deployment.protocolNetwork = this.networkCAIPID
+        })
```
tilacog, line 527: I'm afraid this might return an array of undefined values, since the result of the mapped function is an `undefined` value.

tilacog: I think it can be fixed by returning the deployment in the end:
```suggestion
          deployment.protocolNetwork = this.networkCAIPID
          return deployment
```

tilacog: I think it can be fixed by returning the deployment in the end:
```suggestion
        return transferredDeployments.map((deployment: any) => {
          deployment.protocolNetwork = this.networkCAIPID
          return deployment
        })
```

fordN: yes! good catch, thanks.

packages/indexer-agent/src/agent.ts
```diff
@@ -371,7 +401,65 @@ export class Agent {
...
+                const dayAgo = Math.floor(Date.now() / 1000) - 86400
+                const fourDaysAgo = Math.floor(Date.now() / 1000) - 86400 * 4
+                // L1 deployments being transferred need to be supported for 24hrs post transfer to ensure continued support
+                if (+matchingTransfer.protocolNetwork < 10 && matchingTransfer.transferredToL2At.toNumber() > dayAgo) {
+                    decision.toAllocate = true
+                }
+                // L2 Deployments
+                if (+matchingTransfer.protocolNetwork > 10 && matchingTransfer.transferredToL2At.toNumber() > fourDaysAgo) {
+                    decision.toAllocate = true
+                }
```
tilacog, line 449: I think that the expression `+matchingTransfer.protocolNetwork` will result in a NaN because it is trying to coerce a CAIP2-ID string (eg. `eip155:1`) into a number.

tilacog: I can write a helper function for this kind of operation, such as 
- `isL1("eip155:1") -> true`
- `isL1("eip155:42161") -> false`
- `isL2("eip155:1") -> false`
- `isL2("eip155:42161") -> true`

and use it here.

packages/indexer-agent/src/agent.ts
```diff
@@ -371,7 +401,65 @@ export class Agent {
...
+              }
+            })
+
+            return allocationDecisions
+          },
+        )
+
+        return a
```
tilacog, line 457: I think you meant this, right?
```suggestion
        return allocationDecisions```

tilacog:
Do we need to update the indexing rules to unallocate on L1 after migrating successfully to L2? (and the 24h retention period has passed)?

PR #693, opened 2023-07-27
fix: handle errors fetching actions and vouchers from the DB 

pcarranzav:
This fix adds error handling to the two timers that periodically query
the database: one which gets actions from the queue, and other that gets pending
vouchers. In tests, the DB tables are dropped, so the timers would throw
errors and cause tests to fail. In production whatever errors cause
a DB fetch to fail could be recoverable so the timer will keep retrying
rather than bubbling up an exception, so hopefully that is okay.

We also change CI to only build using Ubuntu 22.04
(the node 18 version has dependency issues on Ubuntu 20.04) and on
node 16 and 18 (the node 17 version was failing to download node, and is EOL).

Additionally we make CI tests and formatting check happen on all PRs, not
only the ones to main.

### DIFF
--- a/packages/indexer-common/src/indexer-management/__tests__/allocations.test.ts
+++ b/packages/indexer-common/src/indexer-management/__tests__/allocations.test.ts
@@ -49,6 +49,7 @@ const setup = async () => {
   sequelize = await connectDatabase(__DATABASE__)
   managementModels = defineIndexerManagementModels(sequelize)
   queryFeeModels = defineQueryFeeModels(sequelize)
+  sequelize = await sequelize.sync({ force: true })
 
   const graphNode = new GraphNode(
     logger,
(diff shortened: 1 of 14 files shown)


packages/indexer-common/src/indexer-management/__tests__/allocations.test.ts
```diff
@@ -99,7 +100,9 @@ const teardownAll = async () => {
 
 describe('Allocation Manager', () => {
   beforeAll(setup)
-  beforeEach(setupEach)
+  beforeEach(() => {
+    return setupEach()
+  })
```
tilacog, line 105: I think the new form is the same as the previous one.
```suggestion
  beforeEach(setupEach)
```

tilacog (approved):
I really liked the failsafe approach.
I just left a comment about a possible revert, but it's all good!
Thanks!

packages/indexer-common/src/indexer-management/__tests__/allocations.test.ts
pcarranzav: ah, yes, sorry, I did this change as I was playing around and missed it in the cleanup!

PR #723, opened 2023-08-15
Fix ci pipeline and update OS dependencies on Ubuntu

DaMandal0rian:
Thie PR attempts to update the OS packages on ubuntu OS for dependencies used by builds such as glibc and fix nodejs/rust bindings.


kapoorabhishek24 (approved):
LGTM

PR #721, opened 2023-08-15
indexer-agent: Various bug fixes and improvements to error handling

fordN:
- Improve graph-node connection testing at startup: use exponential retry and exit process if connection fails. 
- Relax strictness of local deployment and validation of network subgraph (the local deployment may not be needed)
- Improve failure mode if network registration fails or ensuring global index rule fails: log critical message and exit instead of entering unhandled promise rejection.
- Add `protocolNetwork` to action update constraint.
- Fix "try...catch" block setup around staking.channelDisputeEpochs().

PR #724, opened 2023-08-15
Fix contract interface errors

tilacog:
based on #721


pcarranzav (approved):

tilacog:
force pushed to squash commits

PR #725, opened 2023-08-16
fix: use the new Staking ABI when filtering events (hack)

pcarranzav:


tilacog (approved):
Approving to merge and continue the work on the target branch.

PR #735, opened 2023-08-20
fix: properly skip reconciling allocations on a network in manual mode

pcarranzav:
Setting allocation decisions to false would try to close allocations when on manual mode. It wouldn't actually go through because actions won't really be queued on manual, thankfully.
This PR changes it so that we use an empty array of allocation decisions for the network that is in manual mode, making the reconcile step a noop instead of an attempt to close all allocations.


tilacog (approved):
This is really clever and spares the Agent from using resources. 

I'm unsure if other parts of the reconciliation loop require the information for the (denied) allocation decisions, but at least they are logged to the user.

PR #739, opened 2023-08-21
agent: Fix indexer registration

tilacog:
Indexer registration was happening regardless of the indexer being detected on chain.

### DIFF
--- a/packages/indexer-common/src/network.ts
+++ b/packages/indexer-common/src/network.ts
@@ -347,10 +347,10 @@ export class Network {
             ) {
               if (await this.transactionManager.isOperator.value()) {
                 logger.info(`Indexer already registered, operator status already granted`)
-                return
               } else {
                 logger.info(`Indexer already registered, operator status not yet granted`)
               }
+              return
             }
           }
           const receipt = await this.transactionManager.executeTransaction(


pcarranzav (approved):

packages/indexer-common/src/network.ts
```diff
@@ -347,11 +347,11 @@ export class Network {
             ) {
               if (await this.transactionManager.isOperator.value()) {
                 logger.info(`Indexer already registered, operator status already granted`)
-                return
               } else {
                 logger.info(`Indexer already registered, operator status not yet granted`)
               }
```
pcarranzav, line 352: Shouldn't the return go here, so that if the url/geohash doesn't match, it continues with registration?

pcarranzav (changes requested):

packages/indexer-common/src/network.ts
fordN: Or under line 352?  Should avoid sending the registration tx if: already registered, url unchanged, and geoHash unchanged regardless of the operator status.

pcarranzav: Ah yeah that's what I meant, should've commented one line below 👍

pcarranzav (approved):

tilacog:
Will squash commits

PR #741, opened 2023-08-21
Fix string interpolation in migration file

tilacog:
Constraint name was being created as a literal string.
No data is in risk, but the constraint name must be corrected.


pcarranzav (approved):
I suppose this will still leave behind incorrectly named constraints for those that have already run migrations, right? I think we might be able to fix that with an additional migration but it could be done in a separate PR

PR #746, opened 2023-08-22
Add missing fields when parsing network spec from cli args

tilacog:


Theodus (approved):
LGTM. Fixes my issue with opening an allocation to the network subgraph on the local-network

PR #749, opened 2023-08-23
fix(indexer-native): abandon neon-utils in favor of napi-6

Theodus:
I was previously unable to modify indexer-native without the native module failing to self-register. Switching to napi-6 seems to resolve this issue.


Theodus:
Looks like we need to release the new version to use it on the other packages

tilacog:
Rebased on top of latest `main`

dwerner:
Stole this for #975, thanks!

PR #753, opened 2023-08-25
Fix floating promises

tilacog:
Adds `await` before `async` function calls.
- Fixes network subgraph deployments being removed right after they are synced.
- Add logs inside `reconcileDeployment` function.

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -990,6 +990,7 @@ export class Agent {
     targetDeployments: SubgraphDeploymentID[],
     eligibleAllocations: Allocation[],
   ): Promise<void> {
+    const logger = this.logger.child({ function: 'reconcileDeployments' })
     // ----------------------------------------------------------------------------------------
     // Ensure the network subgraph deployment is _always_ indexed
     // ----------------------------------------------------------------------------------------
@@ -997,6 +998,9 @@ export class Agent {
       if (network.networkSubgraph.deployment) {
         const networkDeploymentID = network.networkSubgraph.deployment.id
         if (!deploymentInList(targetDeployments, networkDeploymentID)) {
+          logger.trace('Ensuring Network Subgraph is indexed', {
+            networkDeploymentID,
+          })
           targetDeployments.push(networkDeploymentID)
         }
       }
@@ -1008,6 +1012,9 @@ export class Agent {
     // Ensure all subgraphs in offchain subgraphs list are _always_ indexed
     for (const offchainSubgraph of this.offchainSubgraphs) {
       if (!deploymentInList(targetDeployments, offchainSubgraph)) {
+        logger.trace('Ensuring offchain subgraph is indexed', {
+          offchainSubgraph,
+        })
         targetDeployments.push(offchainSubgraph)
       }
     }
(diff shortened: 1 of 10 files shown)


packages/indexer-agent/src/agent.ts
```diff
@@ -1008,6 +1012,9 @@ export class Agent {
     // Ensure all subgraphs in offchain subgraphs list are _always_ indexed
     for (const offchainSubgraph of this.offchainSubgraphs) {
       if (!deploymentInList(targetDeployments, offchainSubgraph)) {
+        logger.trace('Ensuring offchain subgraph is indexed', {
```
fordN, line 1015: This could potentially emit A LOT of log messages.  I'd rather keep the logs somewhat parsable.  
What about a single log that includes an array of subgraph ids?

packages/indexer-agent/src/agent.ts
```diff
@@ -1020,21 +1027,13 @@ export class Agent {
...
-      // TODO: Return early in here case reconciliation is not needed
-      this.logger.debug('Reconcile deployments', {
-        syncing: activeDeployments.map(id => id.display),
-        target: targetDeployments.map(id => id.display),
-        withActiveOrRecentlyClosedAllocation: eligibleAllocationDeployments.map(
-          id => id.display,
-        ),
-      })
-    }
+    logger.debug('Reconcile deployments', {
```
fordN, line 1030: Removing the conditional?

fordN (changes requested):

packages/indexer-agent/src/agent.ts
```diff
@@ -990,13 +990,17 @@ export class Agent {
...
+    const logger = this.logger.child({ function: 'reconcileDeployments' })
     // ----------------------------------------------------------------------------------------
     // Ensure the network subgraph deployment is _always_ indexed
     // ----------------------------------------------------------------------------------------
-    this.multiNetworks.map(async ({ network }) => {
+    await this.multiNetworks.map(async ({ network }) => {
       if (network.networkSubgraph.deployment) {
         const networkDeploymentID = network.networkSubgraph.deployment.id
         if (!deploymentInList(targetDeployments, networkDeploymentID)) {
+          logger.trace('Ensuring Network Subgraph is indexed', {
```
fordN, line 1001: Do we really want to log this on every loop?

tilacog: I think we can include it as a boolean field in another existing log, like the child log for this function or the `Reconcile deployments` log.

packages/indexer-agent/src/agent.ts
tilacog: This conditional is used only for logging the inputs for deployment reconciliation. 

Having this information in every loop, even when there's no operation to perform, could be helpful as a way of providing evidence.

tilacog: If we remove the conditional, we could merge this log with the following one.
```

File: packages/indexer-agent/src/agent.ts (hunk 1 above)
Comment (tilacog) [reply packages/indexer-agent/src/agent.ts:1015]: I think we can remove this, since offchain deployments are already logged in the `target` field of following logs

File: packages/indexer-agent/src/agent.ts [hunk 4]
```diff
@@ -1046,16 +1034,19 @@ export class Agent {
...
+        target: targetDeployments.map(id => id.display),
+        withActiveOrRecentlyClosedAllocation: eligibleAllocationDeployments.map(
+          id => id.display,
+        ),
         deploy: deploy.map(id => id.display),
         remove: remove.map(id => id.display),
       })
     } else {
-      this.logger.debug('No deployment changes are necessary')
+      logger.debug('No deployment changes are necessary')
```
tilacog, line 1049: @fordN, I've merged the logs, maintaining the same condition: it will only be logged if deployment changes occur.

packages/indexer-agent/src/agent.ts
tilacog: Resolving this thread in favor of https://github.com/graphprotocol/indexer/pull/753#discussion_r1306105071

tilacog:
I'll squash all commits before merging.

fordN (approved):
👍

PR #759, opened 2023-08-30
Relax network subgraph validation in Indexer Service startup

tilacog:
Emit warnings instead of throwing an error, which is causing the process to exit.


hopeyen (approved):
lgtm 👍

PR #762, opened 2023-09-04
Agent: warn and wait if subgraph is behind the chain head

tilacog:


tilacog:
CI results differ between the two previous runs: failed tests in one run have passed in the other.

tilacog:
Rebased on top of latest `main`

packages/indexer-agent/src/commands/start.ts
```diff
@@ -359,6 +379,36 @@ export async function createNetworkSpecification(
   const chainId = await fetchChainId(networkProvider.url)
   const networkIdentifier = resolveChainId(chainId)
 
+  // Warn about inappropriate max block distance for subgraph threshold checks for given networks.
```
hopeyen, line 382: why are the warnings for Arbitrum only?

packages/indexer-agent/src/commands/start.ts
```diff
@@ -37,6 +37,9 @@ import { displayZodParsingError } from './error-handling'
 // eslint-disable-next-line  @typescript-eslint/no-explicit-any
 export type AgentOptions = { [key: string]: any } & Argv['argv']
 
+const DEFAULT_SUBGRAPH_MAX_BLOCK_DISTANCE = 0
```
hopeyen, line 40: 0 block distance sounds good, just wondering if it could be loosened to 2 or 5 or so for the default behavior? for subgraphs deployed locally, what about using `synced` from the indexing statuses query instead of calculating the block number freshness?

packages/indexer-common/src/subgraphs.ts
```diff
@@ -323,3 +327,162 @@ export function isDeploymentWorthAllocatingTowards(
...
+        query: print(updatedQuery),
+      })
+      throw new Error(errorMsg)
+    }
+
+    // Obtain the latest network block number and subgraph query in parallel.
+    const subgraphQueryPromise = subgraph.query(updatedQuery, variables) as Promise<
+      QueryResult<Data> & BlockNumberInterface
+    >
+    const latestNetworkBLockPromise = this.provider.getBlockNumber()
```
hopeyen, line 437: `latestNetworkBLockPromise` cases to `latestNetworkBlockPromise`?

packages/indexer-service/src/commands/start.ts
```diff
@@ -287,6 +304,56 @@ export default {
...
+      }
+    }
+
+    const subgraphFreshnessChecker = new SubgraphFreshnessChecker(
+      'Network Subgraph',
+      networkProvider,
+      argv.subgraphMaxBlockDistance,
+      argv.subgraphFreshnessSleepMilliseconds,
+      logger.child({ component: 'FreshnessChecker' }),
+      Infinity,
```
hopeyen, line 354: is it possible to get stuck in an infinite retry loop for a Network subgraph?

packages/indexer-common/src/subgraphs.ts
```diff
@@ -323,3 +327,162 @@ export function isDeploymentWorthAllocatingTowards(
...
+    let updatedQuery = query
+    try {
+      updatedQuery = mergeSelectionSets(query, blockNumberQuery)
+    } catch (err) {
+      const errorMsg = `Failed to append block number into ${this.subgraphName} query`
+      this.logger.error(errorMsg, { subgraph: this.subgraphName, query: print(query) })
+      throw new Error(errorMsg)
+    }
+
+    // Try obtaining a fresh subgraph query at most 10 times
```
hopeyen, line 408: at most `this.retries` times?

packages/indexer-agent/src/commands/start.ts
tilacog: Default values were tested and set for Ethereum networks (Mainnet & Goerli), but Arbitrum has a faster block speed, so the default config will most likely stall the Agent if used on Arbitrum networks.

packages/indexer-agent/src/commands/start.ts
tilacog: Based on our manual tests, Ethereum protocol subgraphs were synced most of the time (block distance = 0), so we didn't set the default to a higher value. 

I think we could use the `synced` attribute, but we would need to add an extra call to Graph Node just to check the synced status. With the current design, we keep the same number of calls (graphql queries, in this case) as before.

I'm open to discussing the benefits of either approach.

packages/indexer-common/src/subgraphs.ts
tilacog: Thanks. Fixed in c0fe44e2

packages/indexer-common/src/subgraphs.ts
tilacog: I'm not sure I understand the question.

hopeyen: just a subtle difference in casing: `BLock` to `Block`. I don't know if there's a convention to use `L` instead of `l` in `block`

packages/indexer-agent/src/commands/start.ts
hopeyen: Makes sense, thank you, this is more of a curiosity question than anything

To me, I can see that
Graph node status approach 
- Pros: network agnostic, no need for block calculation and so no need for max_block_distance config, graph node and indexer would be on the same page with `sync` status abstracting away the rpc providers
- Cons: additional call to graph node, may affect query service performance

Current design
- Pros: keep the same number of graphql queries made to Graph node, control of freshness requirement falls to indexer instead of graph node status calculation
- Cons: may require more granular control over max_block_instance configuration for specific networks, require indexers to understand freshness requirement 

I think either way is good and can be discussed in a later issue

packages/indexer-service/src/commands/start.ts
tilacog: Yes, and that was a requirement, as aligned with @fordN.

We added a retry count mainly for code testability. We don't plan to show the retry count in the configuration.

packages/indexer-agent/src/commands/start.ts
```diff
@@ -359,6 +379,36 @@ export async function createNetworkSpecification(
   const chainId = await fetchChainId(networkProvider.url)
   const networkIdentifier = resolveChainId(chainId)
 
+  // Warn about inappropriate max block distance for subgraph threshold checks for given networks.
+  if (networkIdentifier.startsWith('eip155:42161')) {
+    // Arbitrum-One and Arbitrum-Goerli
+    if (subgraphs.maxBlockDistance <= DEFAULT_SUBGRAPH_MAX_BLOCK_DISTANCE) {
```
fordN, line 385: This conditional will only get hit if `maxBlockDistance` is negative?

fordN: Should we use something bigger like `50`?

packages/indexer-common/src/subgraphs.ts
```diff
@@ -323,3 +327,162 @@ export function isDeploymentWorthAllocatingTowards(
...
+      latestNetworkBlock,
+      blockDistance,
+      freshnessThreshold: this.threshold,
+      subgraph: this.subgraphName,
+      retriesLeft,
+    }
+    this.logger.trace('Performing subgraph freshness check', logInfo)
+
+    if (blockDistance < 0) {
+      // Invariant violated: Subgraph can't be ahead of network latest block
```
fordN, line 469: If the subgraph is ahead of the provider I think that's ok. SUPER FRESH!
Let's just warn here?

fordN (approved):
I'd just warn in the case of the subgraph being ahead of the provider.  Otherwise looks great! Feel free to merge once you update that part.

PR #768, opened 2023-09-06
Make it work

ArtSuslov:


fordN:
Hey @ArtSuslov 👋 

Without any comment or description I can only make assumptions about the intent of this PR. 

I'm always open to improving usability of the docker containers, etc... but there are a bunch of changes here that I don't see helping the primary users, Graph network indexers. 
- we will not be relying on '@tokene-q/common-ts' instead of `@graphrotocol/common-ts`, are there differences there that you require?
- we are hoping to upgrade our docker images to use a newer version of the Node runtime.  Version 14 is quite old now and even the version we are using, 16, is no longer technically supported.  

I like that you included some instructions for interacting directly with the smart contracts without relying on a browser app.

fordN:
It's been almost a year with no response or activity from the submitter. I'm closing this.

Issue #713, opened 2023-08-11
Add a deployment management mode configuration similar to allocation management mode


fordN:
## Background
Indexers may have other systems/tools for managing deployments, so we'd like to provide the option for turning off deployment management in the indexer-agent.

## Goal
Add feature to support different deployment management modes, so indexers can turn off deployment management. 
Ultimately we can work towards a setup very similar to the allocation management modes: `auto`, `manual` and `oversight` that will give indexers queue management control of deployment changes.  However, in the beginning we can keep it simple and just support `auto` and `manual` modes.  

## Tasks
- [x] Add a startup parameter, `deploymentManagement`, with supported options: `auto` and `manual`
- [x] Add a conditional checking this parameter that will skip reconciling deployments if mode = `manual`

fordN:
## Background
Indexers may have other systems/tools for managing deployments, so we'd like to provide the option for turning off deployment management in the indexer-agent.

## Goal
Add feature to support different deployment management modes, so indexers can turn off deployment management. 
Ultimately we can work towards a setup very similar to the allocation management modes: `auto`, `manual` and `oversight` that will give indexers queue management control of deployment changes.  However, in the beginning we can keep it simple and just support `auto` and `manual` modes.  

## Tasks
- [ ] Add a startup parameter for deployment management mode with supported options: `auto` and `manual`
- [ ] Add a conditional checking this parameter that will skip reconciling deployments if mode = `manual`

PR #769, opened 2023-09-06
Auto and Manual deployment management modes

tilacog:
Resolves https://github.com/graphprotocol/indexer/issues/713

### DIFF
--- a/packages/indexer-common/src/types.ts
+++ b/packages/indexer-common/src/types.ts
@@ -7,6 +7,11 @@ export enum AllocationManagementMode {
   OVERSIGHT = 'oversight',
 }
 
+export enum DeploymentManagementMode {
+  AUTO = 'auto',
+  MANUAL = 'manual',
+}
+
 export enum OrderDirection {
   ASC = 'asc',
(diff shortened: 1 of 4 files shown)


packages/indexer-common/src/types.ts
```diff
@@ -90,3 +95,14 @@ export interface TransactionConfig extends providers.TransactionRequest {
   gasBump: BigNumber
   type: TransactionType
 }
+
+export function parseDeploymentManagementMode(input: string): DeploymentManagementMode {
```
fordN, line 99: For allocation management mode we do something like this to parse the input to the enum variant value: 

```
const allocationManagementMode =
      AllocationManagementMode[
        argv.allocationManagement.toUpperCase() as keyof typeof AllocationManagementMode
      ]
```     
This approach may be a bit more flexible to support adding enum variants without needing to change the parsing function.  What do you think?

fordN (commented):
Looks solid! Just a question about how we parse the yarg.

fordN (approved):
👍

PR #770, opened 2023-09-07
Update arbitrum-one.md

fattox:
From my understanding, this is the correct endpoint for Arbitrum. In addition, i believe this variable is also planned to change, with the old one being deprecated.


fattox:
Payne has mentioned that `.network` is deprecated, still works for me, but feel free to remove that also if it's the case.

fordN:
Closing in favor of #953.  Thanks for your contribution, @fattox.

PR #798, opened 2023-10-09
Refine subgraph query result shape validation

tilacog:
- **Hardening**: refine subgraph query result shape validation
- **Bugfix**: account for error property named `errors` and `error`
- **Bugfix**: pass query `variables` value when recursing


hopeyen (approved):
🏁

PR #805, opened 2023-10-17
Conform grafting auto-resolve feature to latest `main`

tilacog:


fordN:
We may revive this later, but ideally this will be solved in graph-node.

PR #818, opened 2023-11-14
feat: optionally override address book

Theodus:
depends on https://github.com/graphprotocol/common-ts/pull/114


Jannis (approved):
Generally cool with this, but see https://github.com/graphprotocol/common-ts/pull/114#pullrequestreview-1729918619.

fordN (changes requested):
To work with the current latest `common-ts` we'll need to add functionality to read address book from file if user provides a file location

gusinacio (approved):

PR #825, opened 2023-11-22
indexer-agent: Improve robustness of DAI contract contract calls

fordN:
### Changes
- Add fallback if decimals() call to DAI contract fails (default to 18)


abarmat (approved):

kapoorabhishek24 (approved):
LGTM

PR #838, opened 2024-01-04
agent: if alias does not exist try the actual flag

saihaj:
when trying to run locally for some reason `yargs` doesn't show me `dir` as a parsed arg, I would expect it to create aliases automatically so we can reference it. But adding this as safe guards for anyone it doesn't work.


fordN (approved):
LGTM 👍

Issue #803, opened 2023-10-13
Refine subgraph freshness check logs


alex-pakalniskis:
## Current behavior
* Relevant logs do not specify whether the local subgraph deployment or external endpoint are behind, making it difficult to debug a subgraph deemed "not fresh".  
    * The subgraph query will switch between local deployment and external endpoint depending on which is further behind. 

## Preferred behavior
* Logs will indicate whether "not fresh" warning is caused by local subgraph deployment or external endpoint.

PR #839, opened 2024-01-08
common: log endpoint in freshness checker

saihaj:
Today indexers do not have a way to know what subgraph is "not fresh" this PR aims to introduce the `endpoint` of subgraph we consider is "not fresh" in the freshness checker

closes https://github.com/graphprotocol/indexer/issues/803

### DIFF
--- a/packages/indexer-common/src/network-subgraph.ts
+++ b/packages/indexer-common/src/network-subgraph.ts
@@ -44,6 +44,11 @@ export class NetworkSubgraph {
   logger: Logger
   freshnessChecker: SubgraphFreshnessChecker | undefined
   endpointClient?: AxiosInstance
+  /** Endpoint URL for the Network Subgraph Endpoint from the config  */
+  private networkSubgraphConfigEndpoint?: string
+  /** Endpoint URL for the Network Subgraph Endpoint from the deployment  */
+  private networkSubgraphDeploymentEndpoint?: string
+  endpoint?: string
 
   public readonly deployment?: {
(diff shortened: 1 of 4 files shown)


fordN:
Hey sorry forgot to detail this before you submitted PR. 

Would you mind altering commit format a bit to be consistent with past contributions on this repo? 
`<package_name>: <Commit description>`

Lately I've been shortening the package name to leave more room for description, so using `common` instead of `indexer-common`. `*` for something that applies to all packages.

packages/indexer-common/src/network-subgraph.ts
```diff
@@ -54,6 +55,10 @@ export class NetworkSubgraph {
   private constructor(options: NetworkSubgraphOptions) {
     this.logger = options.logger
     this.freshnessChecker = options.subgraphFreshnessChecker
+    this.endpoint =
```
fordN, line 58: This will give us a static endpoint value depending on the config, right? 
The client is dynamic though depending on if the local deployment is synced to chain head (switches to the external endpoint if the local deployment is not synced).

saihaj: how do you know what is an external endpoint? I saw that this class either calls the `options.endpoint` or uses the `deployment` subgraph (which I assume is what local deployed to graph-node)

fordN: options.endpoint is what I mean by "external".  Though you are right it doesn't necessarily mean it's external, it's just a url.

PR #841, opened 2024-01-10
feat(indexer-service): Add graph-node version endpoint from status api

abarmat:
The graph-node is exposing the version information using the following struct, however, it's not exposed through the indexer-service status endpoint. 

This PR adds that information so that the gateway can have access to it.

```
struct Version {
    version: String,
    commit: String,
}
```

### DIFF
--- a/packages/indexer-service/src/server/status.ts
+++ b/packages/indexer-service/src/server/status.ts
@@ -34,7 +34,7 @@ export const createStatusServer = async ({
     'cachedEthereumCalls',
     'subgraphFeatures',
     'apiVersions',
-    'versions',
+    'version',
   ]
   const filteredSchema = wrapSchema({
     schema,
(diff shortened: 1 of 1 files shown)


packages/indexer-service/src/server/status.ts
```diff
@@ -34,6 +34,7 @@ export const createStatusServer = async ({
     'cachedEthereumCalls',
     'subgraphFeatures',
     'apiVersions',
+    'versions',
```
Theodus, line 37:
```suggestion
    'version',
```

Theodus (changes requested):

Theodus (approved):
Ideally this would be surfaced via `/version`. But it's more important that this gets to indexers ASAP so that we can start enforcing a graph-node feature set.

fordN (approved):
👍

PR #842, opened 2024-01-10
*: Downgrade lerna dependency

fordN:
- To be upgraded again once Dockerfile and workflows are upgraded. Version lerna@8.0.2 was not compatible with Node v16.


saihaj (approved):

saihaj:
we should eventually remove `lerna` and just utilize `yarn workspaces` for simplicity

PR #844, opened 2024-01-11
agent,service: improve exception handling

saihaj:
Introducing `uncaughtException` and `unhandledRejection` handlers and gracefully shutdown the process.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,6 @@
     "packages/*"
   ],
   "npmClient": "yarn",
+  "useWorkspaces": true,
   "version": "0.21.0"
 }
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
   ],
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "0.21.0"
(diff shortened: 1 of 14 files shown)


fordN:
We already have similar code paths in the `/commands/start.ts` files of both `indexer-service` and `indexer-agent`.  I see that this is one level up in the index.ts files, but I wonder if there is opportunity to unify the approaches?

fordN (commented):
We already have similar code paths in the `/commands/start.ts` files of both [indexer-service](https://github.com/graphprotocol/indexer/blob/5a743821ab19c316ab064cb7794a36bf32a3860a/packages/indexer-service/src/commands/start.ts#L211-L221) and [indexer-agent](https://github.com/graphprotocol/indexer/blob/5a743821ab19c316ab064cb7794a36bf32a3860a/packages/indexer-agent/src/commands/start.ts#L444-L455).  I see that this is one level up in the index.ts files, but I wonder if there is opportunity to unify the approaches?

saihaj:
yeah you are right we don't then technically need cause that is already picking up from the process

PR #862, opened 2024-02-29
fix: paginate allocation queries

Theodus:
Previously, the indexer-service would fail to handle queries if it had more than 1000 active allocations.


abarmat (approved):
🌅

PR #864, opened 2024-03-01
fix: rollback cost-model to v0.1.16

Theodus:
This gets us back to a situation where we can build & test the main branch on the local network.


dwerner:
Spoke with @Theodus - he's using prepared docker images now and this can be closed.

PR #868, opened 2024-03-08
agent,common: Migrate subgraph deployment assignments to use explicit 'paused' mechanism

fordN:
Graph-node recently added support for explicit pausing of subgraph deployments (in the past they were paused virtually by assigning the deployment to a nonexistent node). 

This PR adds a migration that interacts with the Graph Node indexingStatuses query and JSON-RPC server to update all virtually paused deployments (node = 'removed') to use the new explicit pausing mechanism. It also updates the process of deploying subgraphs, the indexer-agent will no longer specify the index node to which the deployment is assigned; instead the graph-node makes the assignment decisions. 

## Changes
- Migrate subgraph deployment assignments to use explicit pause mechanism
- Use recently added subgraph_pause method to explicitly pause subgraph deployments rather than virtually pausing them by assigning to a non-existent index node
- Remove index-node-ids startup parameter from indexer-agent
- Never send an index-node-id param with subgraph_deploy requests, the receiving graph-node will make the index-node-id decision based on the graph-node.toml config file

### DIFF
--- a/packages/indexer-service/src/server/__tests__/server.test.ts
+++ b/packages/indexer-service/src/server/__tests__/server.test.ts
@@ -64,10 +64,7 @@ const setup = async () => {
   contracts = await connectContracts(getTestProvider('sepolia'), 11155111, undefined)
   sequelize = await sequelize.sync({ force: true })
   const statusEndpoint = 'http://127.0.0.1:8030/graphql'
-  indexingStatusResolver = new IndexingStatusResolver({
-    logger: logger,
-    statusEndpoint,
-  })
+  const queryEndpoint = 'http://127.0.0.1:8000/'
 
   const INDEXER_TEST_API_KEY: string = process.env['INDEXER_TEST_API_KEY'] || ''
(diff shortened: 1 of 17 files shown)


dwerner (approved):
lgtm

packages/indexer-service/src/server/__tests__/server.test.ts
```diff
@@ -64,18 +64,14 @@ const setup = async () => {
   contracts = await connectContracts(getTestProvider('sepolia'), 11155111, undefined)
   sequelize = await sequelize.sync({ force: true })
   const statusEndpoint = 'http://127.0.0.1:8030/graphql'
-  indexingStatusResolver = new IndexingStatusResolver({
-    logger: logger,
-    statusEndpoint,
-  })
+  const queryEndpoint = 'http://127.0.0.1:8000/'
```
fordN, line 67: I think this got clobbered when we were doing the rebase the other day @dwerner. Taking the opportunity to fixup here.

dwerner (approved):
lgtm v2

PR #831, opened 2023-12-07
collect tap ravs

hopeyen:
- Import escrow contracts [npm package](https://www.npmjs.com/package/@semiotic-labs/tap-contracts-bindings)
- Add sequelize schema and resolvers for `scalar_tap_ravs`; updated relations between `Voucher` and `AllocationSummary` to be compatible with `RAV`. 
- Define `SignedRAV` as the equivalence of `TAPVerifier.SignedRAVStruct == EIP712SignedMessage<ReceiptAggregateVoucher>`
- RAV processing triggered alongside `AllocationReceiptCollector`'s voucher processing which is in `indexer-agent`. 
- Add `EventualAllocations` and `EscrowContracts` to `AllocationReceiptCollector`
- Obtain typed final `SignedRAVs` from DB, loop through individual `SignedRav` for individual submissions on-chain
- Refactored `allocationSigner` function to provide `Signer` for computing `allocationIDProof` as required by `escrow.redeem`
- `transactionManager` to first estimate gas and then execute transaction to `contracts.escrow.redeem(signedRAV, allocationIDProof)`
- Upon getting transaction receipt, go ahead deleting RAV from DB.
- Basic metrics added for `RAV`

PR #869, opened 2024-03-14
Collect tap ravs

carlosvdr:
A continuation of PR [831](https://github.com/graphprotocol/indexer/pull/831)
Were we fix redeeming for ravs, collection of allocations, unit tests, migrations among others


packages/indexer-common/src/query-fees/object-conversion.ts
```diff
@@ -0,0 +1,42 @@
+import _ from 'lodash'
```
gusinacio, line 1: We can remove this code, I think it's not being used.

packages/indexer-common/package.json
```diff
@@ -44,6 +45,9 @@
     "lodash.groupby": "^4.6.0",
     "lodash.isequal": "4.5.0",
     "lodash.xor": "^4.5.0",
+    "lodash.transform": "^4.6.0",
+    "lodash.camelcase": "^4.3.0",
+    "lodash.snakecase": "^4.1.1",
```
gusinacio, line 50: This is not used anymore.

packages/indexer-common/src/query-fees/models.ts
```diff
@@ -1,7 +1,8 @@
-import { DataTypes, Sequelize, Model, Association } from 'sequelize'
-import { Address } from '@graphprotocol/common-ts'
+import { DataTypes, Sequelize, Model, Association, CreationOptional } from 'sequelize'
+import { Address, toAddress } from '@graphprotocol/common-ts'
 import { caip2IdRegex } from '../parsers'
-
+import { TAPVerifier } from '@semiotic-labs/tap-contracts-bindings'
+// import { JSONParse, JSONStringify, camelize, snakefy } from './object-conversion'
```
gusinacio, line 5: remove this comment

packages/indexer-common/src/query-fees/models.ts
```diff
@@ -47,6 +48,69 @@ export class Voucher extends Model<VoucherAttributes> implements VoucherAttribut
...
+    allocationId: string
+    timestampNs: bigint
+    valueAggregate: bigint
+  }
+  signature: {
+    r: string
+    s: string
+    v: number
+  }
+}*/
```
gusinacio, line 75: Unused comment

packages/indexer-common/src/query-fees/models.ts
```diff
@@ -47,6 +48,69 @@ export class Voucher extends Model<VoucherAttributes> implements VoucherAttribut
...
+export class ReceiptAggregateVoucher
+  extends Model<ReceiptAggregateVoucherAttributes>
+  implements ReceiptAggregateVoucherAttributes
+{
+  public allocation_id!: Address
+  public sender_address!: Address
+  public signature!: Uint8Array
+  public timestamp_ns!: bigint
+  public value_aggregate!: bigint
+  // public rav!: SignedRAV
```
gusinacio, line 86: Remove comment

packages/indexer-common/src/query-fees/models.ts
```diff
@@ -227,6 +296,68 @@ export function defineQueryFeeModels(sequelize: Sequelize): QueryFeeModels {
...
+          const rawValue = this.getDataValue('sender_address')
+          return toAddress(rawValue)
+        },
+        set(value: string) {
+          const addressWithoutPrefix = value.toLowerCase().replace('0x', '')
+          this.setDataValue('sender_address', addressWithoutPrefix)
+        },
+      },
+      signature: {
+        type: DataTypes.BLOB,
```
gusinacio, line 328: try to use "camelCase"

packages/indexer-common/src/allocations/types.ts
```diff
@@ -26,3 +34,143 @@ export enum AllocationStatus {
...
+    currentAllocations: Allocation[],
+  ): Promise<Allocation[]> => {
+    logger.debug('Refresh eligible allocations', {
+      protocolNetwork,
+    })
+
+    try {
+      const currentEpochResult = await networkSubgraph.query(gql`
+        query {
+          graphNetwork(id: "1") {
```
gusinacio, line 65: This doesn't seem to be a "type"

packages/indexer-service/src/allocations.ts
```diff
@@ -195,11 +52,11 @@ export const ensureAttestationSigners = ({
...
 
           // Derive an epoch and subgraph specific private key
-          const signer = allocationSigner(wallet, allocation)
+          const signerPK = allocationSignerPrivateKey(wallet, allocation)
           const nativeSigner = new NativeAttestationSigner(
             chainId,
             disputeManagerAddress,
-            signer,
+            signerPK,
             allocation.subgraphDeployment.id.bytes32,
```
gusinacio, line 60: Why do we need to update indexer-service?

packages/indexer-common/src/query-fees/models.ts
gusinacio: https://sequelize.org/docs/v6/other-topics/naming-strategies/

gusinacio (commented):
This is the first review, I still need to go through tests and query-fees.ts

packages/indexer-common/src/allocations/types.ts
carlosvdr: would you name the file utils? seems more like it for allocations

packages/indexer-common/package.json
carlosvdr: its used in the ravTest

packages/indexer-agent/src/commands/start.ts
```diff
@@ -427,6 +444,7 @@ export async function createNetworkSpecification(
       subgraphs,
       networkProvider,
       addressBook: argv.addressBook,
+      escrowAddressBook,
```
aasseman, line 447: Can't we add the escrow addresses into the original `addressBook` instead of adding all of that code?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -259,6 +301,7 @@ export class AllocationReceiptCollector implements ReceiptCollector {
         // Remove the batch from the processing queue
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const batch = this.receiptsToCollect.pop()!
+        this.logger.info(` ========${batch} ============`)
```
aasseman, line 304: Is this line needed?

packages/indexer-common/src/allocations/types.ts
aasseman: "eventuals" or "monitor"

packages/indexer-agent/src/commands/start.ts
carlosvdr: it was done in this way so that if necesary you can pass the values for the escrowAddress, else it will take those it has already. (this was done overall to make testing easier)

packages/indexer-common/src/allocations/query-fees.ts
carlosvdr: can be removed, probably was added to get more info

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -624,6 +841,126 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        `Undefined escrow contracts, but this shouldn't happen as RAV process is only triggered when escrow is provided`,
+        {
+          signedRavs,
+        },
+      )
+      return
+    }
+    const escrow = this.escrowContracts
+
+    logger.info(`Redeem final RAVs on chain individually`, {
```
aasseman, line 860:
```suggestion
    logger.info(`Redeem allocations' last RAVs on chain individually`, {
```

packages/indexer-common/src/allocations/types.ts
aasseman, line 1: Besides the file naming issue, are you sure that such monitoring isn't already being done somewhere in the code? I found this suspicious import: https://github.com/graphprotocol/indexer/blob/b58372cbf6149df6de59f8df8535bbbaa8224d11/packages/indexer-service/src/commands/start.ts#L26

packages/indexer-common/src/query-fees/models.ts
```diff
@@ -47,6 +48,69 @@ export class Voucher extends Model<VoucherAttributes> implements VoucherAttribut
   }
 }
 
+export interface ReceiptAggregateVoucherAttributes {
+  allocation_id: string
+  sender_address: string
+  signature: Uint8Array
+  timestamp_ns: bigint
+  value_aggregate: bigint
+  //rav: unknown
```
aasseman, line 57: This line should be removed

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,188 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        const Op = Sequelize.Op
+        await this.models.receiptAggregateVouchers.update(
+          { final: true },
+          {
+            where: {
+              allocation_id: { [Op.in]: ravsToFinalize },
+            },
+          },
+        )
+      }
```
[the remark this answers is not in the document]
carlosvdr: Yeah figured there is a way using 
```
await this.models.receiptAggregateVouchers.update(
        { final: true },
        {
          where: {
            allocationId: { [Op.in]: allocationIds },
            redeemedAt: { [Op.lt]: new Date(finalizeTimeFrame) },
          },
        },
      )
```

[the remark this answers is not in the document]
carlosvdr: Yeah figured there is a way using 
```
await this.models.receiptAggregateVouchers.update(
        { final: true },
        {
          where: {
            allocationId: { [Op.in]: allocationIds },
            redeemedAt: { [Op.lt]: new Date(finalizeTimeFrame) },
          },
        },
      )
```

packages/indexer-common/src/allocations/types.ts
carlosvdr: It does seem like it, I will remove it from indexer-common and see if anything breaks

carlosvdr: Looks like both are needed
first in indexer-service to be able to call ensureAttestationSigners
https://github.com/graphprotocol/indexer/blob/b58372cbf6149df6de59f8df8535bbbaa8224d11/packages/indexer-service/src/commands/start.ts#L442-L448

carlosvdr: Second is to create the AllocationReceiptCollector
https://github.com/graphprotocol/indexer/blob/b58372cbf6149df6de59f8df8535bbbaa8224d11/packages/indexer-common/src/network.ts#L260-L270

aasseman: Ok then, thanks for checking!

aasseman: Yes but does `monitorEligibleAllocations` really need to implemented a second time?

carlosvdr: hmmm no monitorning probably not, Ill see if theres a different way to get this and only call it once

gusinacio: If I'm not mistaken, it was moved. It was used only on indexer-service, but now it's needed on indexer-agent as well.
It's not implemented a second time but rather moved from indexer-service.

packages/indexer-common/package.json
gusinacio: The test that uses it should be removed because we are not parsing receipts from the database anymore.

packages/indexer-common/src/query-fees/object-conversion.ts
gusinacio: The whole file is only used on a test, we can remove that test because it's not how receipts are stored in the database anymore

packages/indexer-common/src/query-fees/__test__/rav.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        timestampNs: 18446744073709551605n,
+        valueAggregate: 340282366920938463463374607431768211455n,
+      },
+      signature: {
+        r: '0x56f8e2b7fecee908ff0d28ed172a61bee8ffe87b498c23da7eaca74cd7c5b5ce',
+        s: '0x4ead667e3eb4f1d8c4a20e6f0e5dbe73f9f63e9d0d23e606b3239332c5f21337',
+        v: 27,
+      },
+    })
+  })
```
gusinacio, line 29: This test can be removed because the database doesn't parse json anymore.

packages/indexer-service/src/allocations.ts
gusinacio: It has changed because now allocationSigner was updated to be a `Signer`

packages/indexer-common/src/allocations/types.ts
```diff
@@ -1,28 +0,0 @@
...
-  queryFeesCollected: BigNumber | undefined
-}
-
-export enum AllocationStatus {
-  NULL = 'Null',
-  ACTIVE = 'Active',
-  CLOSED = 'Closed',
-  FINALIZED = 'Finalized',
-  CLAIMED = 'Claimed',
-}
```
gusinacio, line 28: Why was this file removed? I think we should have both `types.ts` and `monitor.ts`

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+  ): Eventual<ValidRavs> {
+    return pendingRAVs.tryMap(
+      async (pendingRAVs) => {
+        return await pReduce(
+          pendingRAVs,
+          async (results, rav) => {
+            // allocationId doesn't have 0x prefix, but still matched in contract calls
+            // destory voucher if AllocationIDTracker.isAllocationIdUsed(rav.allocationId)
+            if (
+              BigNumber.from(rav.rav.rav.valueAggregate).lt(
```
gusinacio, line 511: rav.rav.rav is kinda awkward 😬

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+      )
+
+      // Update those that redeemed_at is older than 20 minutes and mark as final
+      const finalizeTimeFrame = Date.now() - 20 * 60 * 1000
+      const Op = Sequelize.Op
+      await this.models.receiptAggregateVouchers.update(
+        { final: true },
+        {
+          where: {
+            allocationId: { [Op.in]: allocationIds },
```
gusinacio, line 575: You can use  `where: { last: true, final: false,  redeemedAt: { [Op.lt]: new Date(finalizeTimeFrame) }}`

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        )
+      },
+      { onError: (err) => this.logger.error(`Failed to reduce to signed RAVs`, { err }) },
+    )
+  }
+
+  // redeem only if last is true
+  // Later can add order and limit
+  protected async pendingRAVs(): Promise<ReceiptAggregateVoucher[]> {
+    let ravsToRedeem: ReceiptAggregateVoucher[] = []
```
gusinacio, line 531: You can remove here and declare as `const ravsToRedeem` on line 564

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+    const unfinalizedRAVs = await this.models.receiptAggregateVouchers.findAll({
+      where: { last: true, final: false },
+    })
+    // Obtain allocationIds to use as filter in subgraph
+    const allocationIdsAddress = unfinalizedRAVs.map(
+      (rav) => rav.getSignedRAV().rav.allocationId.toLowerCase(),
+    )
+    const allocationIds = unfinalizedRAVs.map(
+      (rav) => rav.getSignedRAV().rav.allocationId.toLowerCase().replace('0x', ''),
+    )
```
gusinacio, line 541: This can be removed

packages/indexer-common/src/allocations/query-fees.ts
gusinacio: Sorry, wrong scope

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+      // From obtained allocationIDs, get those that were not returned and redeem
+      const returnedAllocationIDs = redeemedRavsQueryResp.data.transactions.map(
+        (transaction) => transaction.allocationID,
+      )
+      const nonRedeemedAllocationIDs = allocationIdsAddress.filter(
+        (allocationID) => !returnedAllocationIDs.includes(allocationID),
+      )
+      ravsToRedeem = unfinalizedRAVs.filter((rav) =>
+        nonRedeemedAllocationIDs.includes(rav.getSignedRAV().rav.allocationId),
+      )
```
gusinacio, line 566: In theory, you could just return this below line 579.

packages/indexer-common/src/allocations/query-fees.ts
gusinacio: not sure how to fix tho

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        return ravs
+          .map((rav) => {
+            const domain = {
+              chainId: this.protocolNetwork.split(':')[1],
+              name: 'TAP',
+              version: '1',
+              verifyingContract: this.escrowContracts!.tapVerifier.address,
+            }
+            const signedRav = rav.getSignedRAV()
+            console.log({ domain, rav })
```
gusinacio, line 471: Remove this console.log.

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+                ReceiptAggregateVoucher: [
+                  { name: 'allocationId', type: 'address' },
+                  { name: 'timestampNs', type: 'uint64' },
+                  { name: 'valueAggregate', type: 'uint128' },
+                ],
+              },
+              signedRav.rav,
+              signedRav.signature,
+            )
+            console.log({ signerAddress, ravSigner: rav.senderAddress })
```
gusinacio, line 484: Remove this console.log and maybe convert this to an assert to check if the senderAddress is correct for that allocationId. You may need to find some way to get this information from the subgraph.

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        {
+          where: {
+            allocationId: { [Op.in]: allocationIds },
+            redeemedAt: { [Op.lt]: new Date(finalizeTimeFrame) },
+          },
+        },
+      )
+    }
+    return ravsToRedeem
+  }
```
gusinacio, line 582:
```suggestion
      )
      return unfinalizedRAVs.filter((rav) =>
        nonRedeemedAllocationIDs.includes(rav.getSignedRAV().rav.allocationId),
      )
    }
    return []
  }
```

packages/indexer-common/src/allocations/types.ts
aasseman: Ahhh sorry I got confused I guess :grimacing:

packages/indexer-common/src/allocations/types.ts
aasseman: @carlosvdr yup I made a mistake in the review, I was somehow convinced that this eventual was implemented twice. We need the file back :grimacing:

packages/indexer-common/src/allocations/query-fees.ts
carlosvdr: it already does that

packages/indexer-common/src/allocations/query-fees.ts
carlosvdr: The reason why its done is because in the subgraph is the full address and in the db is without the `0x`, this caused the query not to find those allocations

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+      )
+      const nonRedeemedAllocationIDs = allocationIdsAddress.filter(
+        (allocationID) => !returnedAllocationIDs.includes(allocationID),
+      )
+      ravsToRedeem = unfinalizedRAVs.filter((rav) =>
+        nonRedeemedAllocationIDs.includes(rav.getSignedRAV().rav.allocationId),
+      )
+
+      // Update those that redeemed_at is older than 20 minutes and mark as final
+      const finalizeTimeFrame = Date.now() - 20 * 60 * 1000
```
aasseman, line 569: Can we increase this to 1 hour?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +410,177 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        `,
+        { allocationIdsAddress },
+      )
+      // From obtained allocationIDs, get those that were not returned and redeem
+      const returnedAllocationIDs = redeemedRavsQueryResp.data.transactions.map(
+        (transaction) => transaction.allocationID,
+      )
+      const nonRedeemedAllocationIDs = allocationIdsAddress.filter(
+        (allocationID) => !returnedAllocationIDs.includes(allocationID),
+      )
```
aasseman, line 563: How about resetting `redeemedAt` to `null` for all the `nonRedeemedAllocationIDs`. Just to make sure that the logic is solid in case of a reorg that reverts the redeem?

carlosvdr: Sounds good withe me yes

packages/indexer-common/src/allocations/query-fees.ts
aasseman: I think you don't need to filter by allocationId. Any RAV that remained redeemed on chain for more than N minutes can be marked as final. Even safer if you do https://github.com/graphprotocol/indexer/pull/869#discussion_r1526944344 I think.

gusinacio: It does that, but you are sending all `allocationIds` which is inefficient. You can simplify that by using just last, final and redeemedAt

packages/indexer-common/src/allocations/query-fees.ts
gusinacio: You don't need this `allocationIds` if you use the same where clause in your update.

carlosvdr: I did tried it and it does not work, it should be an exact match per object in the array

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +409,191 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+  // Later can add order and limit
+  protected async pendingRAVs(): Promise<ReceiptAggregateVoucher[]> {
+    const unfinalizedRAVs = await this.models.receiptAggregateVouchers.findAll({
+      where: { last: true, final: false },
+    })
+    // Obtain allocationIds to use as filter in subgraph
+    const allocationIdsAddress = unfinalizedRAVs.map((rav) =>
+      rav.getSignedRAV().rav.allocationId.toLowerCase(),
+    )
+    const allocationIds = unfinalizedRAVs.map((rav) =>
```
aasseman, line 536:
```suggestion
    // Lowercase and remove '0x' prefix of addresses to match format in TAP DB Tables
    const unfinalizedRavsAllocationIdsTrunc = unfinalizedRAVs.map((rav) =>
```

packages/indexer-common/src/allocations/query-fees.ts
aasseman:
```suggestion

    if (allocationIdsAddress.length > 0) {
      console.log(`Get on-chain redeemed RAVs from TAP subgraph`)
```

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +409,191 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+              allocationID
+            }
+          }
+        `,
+        { allocationIdsAddress },
+      )
+      // From obtained allocationIDs, get those that were not returned and redeem
+      const returnedAllocationIDs = redeemedRavsQueryResp.data.transactions.map(
+        (transaction) => transaction.allocationID,
+      )
```
aasseman, line 557: Since you don't use `redeemedRavsQueryResp` for anything else, why not extract the data you need early?
```suggestion
      const chainRedeemedRavAllocationIds = await this.tapSubgraph.query(
        gql`
          query transactions($allocationIdsAddress: [String!]!) {
            transactions(
              where: { type: "redeem", allocationID_in: $allocationIdsAddress }
            ) {
              allocationID
            }
          }
        `,
        { allocationIdsAddress },
      ).data.transactions.map(
        (transaction) => transaction.allocationID,
      )
```

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +409,191 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+      // Subgraph returns addresses wiht 0x prefix
```
aasseman, line 558:
```suggestion
      // Filter unfinalized RAVs from DB to keep the ones not redeemed on-chain
```

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +409,191 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+      const nonRedeemedAllocationIDAddresses = allocationIdsAddress.filter(
+        (allocationID) => !returnedAllocationIDs.includes(allocationID),
+      )
+      const nonRedeemedAllocationIDs = nonRedeemedAllocationIDAddresses.map(
```
aasseman, line 562:
```suggestion
      // Lowercase and remove '0x' prefix of addresses to match format in TAP DB Tables
      const nonRedeemedAllocationIDsTrunc = nonRedeemedAllocationIDAddresses.map(
```

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +409,191 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        (allocationID) => allocationID.toLowerCase().replace('0x', ''),
+      )
+
```
aasseman, line 565:
```suggestion

      // Mark RAVs as unredeemed in DB if the TAP subgraph couldn't find the redeem Tx.
      // To handle a chain reorg that "unredeemed" the RAVs.
```

aasseman (commented):
This round is about the readability of `pendingRAVs`

packages/indexer-common/src/allocations/query-fees.ts
carlosvdr: If run like this it wont let you access `data` due to the object being just a promise it is until afterwards I can access it

aasseman: I think parenthesis around and it'll work:
```ts
(await this.tapSubgraph.query(
        gql`
          query transactions($allocationIdsAddress: [String!]!) {
            transactions(
              where: { type: "redeem", allocationID_in: $allocationIdsAddress }
            ) {
              allocationID
            }
          }
        `,
        { allocationIdsAddress },
      )).data[...]
```

packages/indexer-common/src/query-fees/object-conversion.ts
```diff
@@ -0,0 +1,26 @@
+export const JSONParse = (json: string) => {
```
gusinacio, line 1: Is this file needed?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +412,194 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+        },
+      )
+
+      // Update those that redeemed_at is older than 60 minutes and mark as final
+      const finalizeTimeFrame = Date.now() - this.finalityTime * 1000
+      await this.models.receiptAggregateVouchers.update(
+        { final: true },
+        {
+          where: {
+            allocationId: { [Op.in]: unfinalizedRavsAllocationIdsTrunc },
```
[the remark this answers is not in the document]
carlosvdr: Yeah missed that part in the comments Im adding that

packages/indexer-common/src/query-fees/object-conversion.ts
carlosvdr: think yeah its not needed

gusinacio (approved):
LGTM

packages/indexer-common/src/allocations/__tests__/docker-compose.yaml
aasseman, line 1: No that's not the correct way to do it. You'd let devs figure it out on their own on their own setups.

However you do want to make sure the CI tests will have something like this. It's actually already taken care of here: https://github.com/graphprotocol/indexer/blob/0245f399166a1ca09732992665f1bd0555f36862/.github/workflows/ci.yml#L21

aasseman (changes requested):

aasseman (commented):
Ready for review by maintainers

aasseman (approved):

packages/indexer-common/src/query-fees/__test__/rav.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const result = await queryFeeModels.receiptAggregateVouchers.findAll()
+    // expect([value]).toEqual(result)
+    expect(result).toEqual([
+      expect.objectContaining({
+        allocationId: value.allocationId,
+        final: value.final,
+        last: value.last,
+        senderAddress: value.senderAddress,
+        signature: value.signature,
+        timestampNs: BigInt(value.timestampNs).toString(),
```
gusinacio, line 40: why do you need to convert this to a string? it's supposed to be a big int. Maybe you should update the model with the right conversion

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +412,189 @@ export class AllocationReceiptCollector implements ReceiptCollector {
     })
   }
 
-  private async pendingVouchers(): Promise<Voucher[]> {
+  protected async pendingVouchers(): Promise<Voucher[]> {
```
gusinacio, line 415: this doesn't need to be changed

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +412,189 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+                ReceiptAggregateVoucher: [
+                  { name: 'allocationId', type: 'address' },
+                  { name: 'timestampNs', type: 'uint64' },
+                  { name: 'valueAggregate', type: 'uint128' },
+                ],
+              },
+              signedRav.rav,
+              signedRav.signature,
+            )
+            console.log({ signerAddress, ravSigner: rav.senderAddress })
```
gusinacio, line 485: Remove this console.log

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +412,189 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+    const unfinalizedRAVs = await this.models.receiptAggregateVouchers.findAll({
+      where: { last: true, final: false },
+    })
+    // Obtain allocationIds to use as filter in subgraph
+    const unfinalizedRavsAllocationIds = unfinalizedRAVs.map((rav) =>
+      rav.getSignedRAV().rav.allocationId.toLowerCase(),
+    )
+
+    if (unfinalizedRavsAllocationIds.length > 0) {
+      console.log(`Get on-chain redeemed RAVs from TAP subgraph`)
```
gusinacio, line 541: if you want to log things, use the logger

packages/indexer-common/src/query-fees/models.ts
```diff
@@ -227,6 +279,69 @@ export function defineQueryFeeModels(sequelize: Sequelize): QueryFeeModels {
...
+        allowNull: false,
+        primaryKey: true,
+        get() {
+          const rawValue = this.getDataValue('senderAddress')
+          return toAddress(rawValue)
+        },
+        set(value: string) {
+          const addressWithoutPrefix = value.toLowerCase().replace('0x', '')
+          this.setDataValue('senderAddress', addressWithoutPrefix)
+        },
```
gusinacio, line 308: Maybe you need to do the conversions to bigint using a similar code to this but for timestampNs and aggregateValue

packages/indexer-common/src/query-fees/models.ts
```diff
@@ -47,6 +47,53 @@ export class Voucher extends Model<VoucherAttributes> implements VoucherAttribut
...
+  public readonly allocationSummary?: AllocationSummary
+
+  public static associations: {
+    allocationSummary: Association<ReceiptAggregateVoucher, AllocationSummary>
+  }
+
+  getSignedRAV(): SignedRAV {
+    return {
+      rav: {
+        allocationId: toAddress(this.allocationId),
```
gusinacio, line 86: unneeded conversion to Address

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -0,0 +1,161 @@
...
+        signedRav.signature,
+      )
+
+      expect(signerAddress).toEqual('0x886574712d0ca20C36FD090A594Df7eCa17cd38e')
+    },
+    timeout,
+  ),
+    test('test submitRAVs', async () => {})
+
+  test('test RAV Processing eventual', async () => {})
```
gusinacio, line 160: do you plan to implement this? if not I suggest removing them

gusinacio (commented):
review post rebase

packages/indexer-common/src/allocations/query-fees.ts
carlosvdr: forgot to remove that one

packages/indexer-agent/src/commands/start.ts
```diff
@@ -238,6 +245,16 @@ export const start = {
         type: 'string',
         required: false,
       })
+      .option('tap-address-book', {
+        description: 'TAP contracts address book file path',
+        type: 'string',
+        required: false,
+      })
+      .option('chain-finalize-time', {
+        description: 'The time in seconds that the chain finalizes blocks',
```
Jannis, line 254: Is time the right thing to use here? In other places (graph node for example) we have been using the number of blocks to be mined after a block before that block is considered final. This is because blockchains don't generate new blocks with high precision.

carlosvdr: We based this off info Tomas gave us, basically 

- the sequencer takes less than a second to order and post the tx in arbitrum
- 6-7 mins is the time the sequencer takes to post the tx in ethereum
- ~13 mins , its the finality time in ethereum post merge
So tit should be around 20 mins the finality time
We are defining it as 60 mins just to be sure
Either way ravs will be saved in the db in case we encounter any issue

tkornuta-semiotic: hey @Jannis is this clearing your concerns? can we solve this issue?

fordN (commented):
Not quite done reviewing everything, but figured it's worth submitting the comments I have now so we can start discussion.

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -624,6 +826,126 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+
+    // Redeem RAV one-by-one as no plual version available
+    for (const { rav: signedRav, allocation, sender } of signedRavs) {
+      const { rav } = signedRav
+      const stopTimer = this.metrics.ravsRedeemDuration.startTimer({
+        allocation: rav.allocationId,
+      })
+      try {
+        // compute allocation id proof
+        //
```
fordN, line 857: Format and cleanup comment?

packages/indexer-agent/src/__tests__/indexer.ts
```diff
@@ -170,13 +170,23 @@ const setup = async () => {
...
+        url: 'https://api.thegraph.com/subgraphs/name/graphprotocol/scalar-tap-arbitrum-sepolia',
+      },
     },
     transactionMonitoring: {
       gasIncreaseTimeout: 240000,
       gasIncreaseFactor: 1.2,
       baseFeePerGasMax: 100 * 10 ** 9,
       maxTransactionAttempts: 0,
     },
+    tapAddressBook: {
```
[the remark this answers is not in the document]
carlosvdr: perhaps yes although this is only for the tests

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
```diff
@@ -0,0 +1,241 @@
...
+  const tables = await queryInterface.showAllTables()
+  logger.debug(`Checking if scalar_tap_receipts table exists`, { tables })
+
+  if (tables.includes('scalar_tap_receipts')) {
+    logger.debug(`scalar_tap_receipts already exist, migration not necessary`)
+  } else {
+    logger.info(`Create scalar_tap_receipts`)
+    await queryInterface.createTable('scalar_tap_receipts', {
+      id: {
+        type: DataTypes.BIGINT,
```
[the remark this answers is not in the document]
carlosvdr: Yeah the amount of receipts can get pretty big

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
```diff
@@ -0,0 +1,241 @@
...
+      allowNull: false,
+    },
+    updated_at: {
+      type: DataTypes.DATE,
+      allowNull: false,
+    },
+  })
+
+  logger.info(`Add primary key`)
+  await queryInterface.addConstraint('scalar_tap_ravs', {
```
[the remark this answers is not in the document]
carlosvdr: Actually Im not sure about that , would it affect to add the constraints in the model too ? @aasseman

packages/indexer-agent/src/db/migrations/13-add-scalar-tap-deny-list.ts
```diff
@@ -0,0 +1,77 @@
...
+interface MigrationContext {
+  queryInterface: QueryInterface
+  logger: Logger
+}
+
+interface Context {
+  context: MigrationContext
+}
+
+export async function up({ context }: Context): Promise<void> {
```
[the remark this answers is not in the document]
carlosvdr: It was failing before if we didnt add this , since it was looking for the deny list and not finidng it would crash

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -0,0 +1,159 @@
...
+          signature: rav.signature,
+          timestampNs: rav.timestampNs,
+          valueAggregate: rav.valueAggregate,
+        }),
+      ])
+    },
+    timeout,
+  )
+
+  // test(
```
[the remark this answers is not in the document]
carlosvdr: Yes its the plan, just wasnt in the priorities

packages/indexer-common/src/allocations/keys.ts
```diff
@@ -38,6 +42,11 @@ export const allocationSigner = (wallet: Wallet, allocation: Allocation): string
   )
 }
 
+// Returns allocation signer wallet
+export const allocationSigner = (wallet: Wallet, allocation: Allocation): Signer => {
```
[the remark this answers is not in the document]
carlosvdr: Yeah looks like this was added afterwards, and decided to use the same naming 
One only returns the privateKey and the other the wallet
If you want to change the name its okay, but I think its okay
Since it returns what it is
the allocationSigner and the allocationSignerPrivateKey

packages/indexer-common/src/allocations/monitor.ts
```diff
@@ -0,0 +1,182 @@
...
+  IndexerErrorCode,
+  parseGraphQLAllocation,
+} from '@graphprotocol/indexer-common'
+import { Allocation, MonitorEligibleAllocationsOptions } from './types'
+
+import gql from 'graphql-tag'
+
+import { Eventual, timer } from '@graphprotocol/common-ts'
+
+export const monitorEligibleAllocations = ({
```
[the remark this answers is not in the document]
carlosvdr: yes

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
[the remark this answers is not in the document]
aasseman: The TAP protocol uses `u128` in the receipts. Give us some headroom for future use cases.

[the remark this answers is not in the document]
aasseman: https://github.com/semiotic-ai/timeline-aggregation-protocol/blob/c179dfedee1ed8078b358de3d35f4d051f74c9c1/tap_core/src/receipt/receipt_sol.rs#L27

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +416,167 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
       limit: this.voucherRedemptionMaxBatchSize, // limit the number of vouchers to the max batch size
     })
   }
 
+  //TODO: Consider adding new configurations for RAV similar to vouchers
+  // (voucherRedemptionThreshold, voucherRedemptionMaxBatchSize, voucherRedemptionBatchThreshold)
+  startRAVProcessing() {
+    const notifyAndMapEligible = (signedRavs: ValidRavs) => {
+      if (signedRavs.belowThreshold.length > 0) {
+        const logger = this.logger.child({})
```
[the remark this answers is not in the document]
carlosvdr: Yeah probably was missed at one point, adding info to indicate its from the RavProcessing portion

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +416,167 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+    const unfinalizedRAVs = await this.models.receiptAggregateVouchers.findAll({
+      where: { last: true, final: false },
+    })
+    // Obtain allocationIds to use as filter in subgraph
+    const unfinalizedRavsAllocationIds = unfinalizedRAVs.map((rav) =>
+      rav.getSignedRAV().rav.allocationId.toLowerCase(),
+    )
+
+    if (unfinalizedRavsAllocationIds.length > 0) {
+      const returnedAllocationIDs = (
```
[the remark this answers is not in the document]
carlosvdr: The naming is due to the transaction not being yet final in the chain
Once it is final we ignore it in the filtering for ravs, hence the naming unfinalized

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +416,167 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+  protected getSignedRAVsEventual(
+    pendingRAVs: Eventual<RavWithAllocation[]>,
+  ): Eventual<ValidRavs> {
+    return pendingRAVs.tryMap(
+      async (pendingRAVs) => {
+        return await pReduce(
+          pendingRAVs,
+          async (results, rav) => {
+            // allocationId doesn't have 0x prefix, but still matched in contract calls
+            // destory voucher if AllocationIDTracker.isAllocationIdUsed(rav.allocationId)
```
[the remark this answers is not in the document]
carlosvdr: This comment looks outdated yeah

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -624,6 +826,126 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
   }
 
+  private async submitRAVs(signedRavs: RavWithAllocation[]): Promise<void> {
+    const logger = this.logger.child({
+      function: 'submitRAVs()',
+      signedRavSize: signedRavs.length,
+    })
+    if (!this.tapContracts) {
+      logger.error(
+        `Undefined escrow contracts, but this shouldn't happen as RAV process is only triggered when escrow is provided`,
```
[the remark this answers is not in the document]
carlosvdr: Probably contact us, meaning there is an error with the bindings

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
fordN: I was referring to `returnedAllocationIDs`. The comment references the line above.

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +416,167 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+          ),
+        })
+      }
+      return signedRavs.eligible
+    }
+
+    const pendingRAVs = this.getPendingRAVsEventual()
+    const signedRAVs = this.getSignedRAVsEventual(pendingRAVs)
+    const eligibleRAVs = signedRAVs
+      .map(notifyAndMapEligible)
```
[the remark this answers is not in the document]
gusinacio: The reason we did this is for testing. It's way easier to test `eventuals` if you inject the data by using "data injection" as suggested by the lib. This way we can test the signedRavsEventual logic without needing to provide a database connection.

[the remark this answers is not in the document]
carlosvdr: Althopugh this tests arent there yet we do plan on adding them

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +416,167 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+  protected async pendingVouchers(): Promise<Voucher[]> {
     return this.models.vouchers.findAll({
       where: { protocolNetwork: this.protocolNetwork },
       order: [['amount', 'DESC']], // sorted by highest value to maximise the value of the batch
       limit: this.voucherRedemptionMaxBatchSize, // limit the number of vouchers to the max batch size
     })
   }
 
+  //TODO: Consider adding new configurations for RAV similar to vouchers
+  // (voucherRedemptionThreshold, voucherRedemptionMaxBatchSize, voucherRedemptionBatchThreshold)
```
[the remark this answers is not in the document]
carlosvdr: Thing here is we had to remove this comment since thats what we already do

packages/indexer-common/src/allocations/query-fees.ts
gusinacio: Hey, any update on this?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -480,6 +491,55 @@ export class AllocationReceiptCollector implements ReceiptCollector {
     )
   }
 
+  private async getAllocationfromAllocationId(
```
aasseman, line 494:
```suggestion
  private async getAllocationsfromAllocationIds(
```

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -480,6 +491,55 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+              }
+              allocatedTokens
+              createdAtEpoch
+              createdAtBlockHash
+              closedAtEpoch
+              closedAtEpoch
+              closedAtBlockHash
+              poi
+              queryFeeRebates
+              queryFeesCollected
```
aasseman, line 525: Do we need all of those fields?

packages/indexer-agent/src/commands/start.ts
```diff
@@ -557,7 +586,14 @@ export async function run(
...
     networkSpecifications,
     async (spec: NetworkSpecification) =>
-      Network.create(logger, spec, queryFeeModels, graphNode, metrics),
+      Network.create(
+        logger,
+        spec,
+        queryFeeModels,
+        graphNode,
+        metrics,
+        sequelize,
```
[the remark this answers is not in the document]
fordN: Checking in on this one

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
```diff
@@ -0,0 +1,241 @@
...
+interface MigrationContext {
+  queryInterface: QueryInterface
+  logger: Logger
+}
+
+interface Context {
+  context: MigrationContext
+}
+
+export async function up({ context }: Context): Promise<void> {
```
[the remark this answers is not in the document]
fordN: Following up on this comment, is this table definition in a model? The models should contain entire definition, so it's not split up between migration and model. Also the migration should really only be needed if the user already has data in the DB (not always the case).

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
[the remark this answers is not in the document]
fordN: Following up here, is there a sequelize model for this table?

packages/indexer-agent/src/db/migrations/13-add-scalar-tap-deny-list.ts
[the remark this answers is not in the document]
fordN: Ok, not sure the context there on why it was crashing. 
Is there a sequelize model for this?

packages/indexer-common/package.json
```diff
@@ -54,6 +55,7 @@
     "parsimmon": "^1.18.1",
     "sequelize": "6.33.0",
     "ts-custom-error": "^3.3.1",
+    "tap-contracts": "semiotic-ai/timeline-aggregation-protocol-contracts#ca1cfa653d0d1d209aee48ab021b0e562d2b5688",
```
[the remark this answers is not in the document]
fordN: Explanation?

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -0,0 +1,159 @@
...
+declare const __DATABASE__: any
+declare const __LOG_LEVEL__: never
+let logger: Logger
+let receiptCollector: AllocationReceiptCollector
+let metrics: Metrics
+let queryFeeModels: QueryFeeModels
+let sequelize: Sequelize
+const timeout = 30000
+
+// const startRAVProcessing = jest.spyOn(
```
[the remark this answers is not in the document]
fordN: Still commented out code here, is there a reason to keep it?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -883,10 +883,15 @@ export class AllocationReceiptCollector implements ReceiptCollector {
 
         try {
           const addressWithoutPrefix = rav.allocationId.toLowerCase().replace('0x', '')
-          await this.models.receiptAggregateVouchers.update(
-            { redeemedAt: new Date() },
-            { where: { allocationId: addressWithoutPrefix } },
-          )
+          // WE use sql directly due to a bug in sequelize update:
```
fordN, line 943: Did this change since my last review?

packages/indexer-common/src/allocations/__tests__/tap.test.ts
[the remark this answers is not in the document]
fordN: Following up here, still see a bunch of commented out code lines...

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  await sequelize.drop({})
+}
+
+describe('TAP', () => {
+  beforeAll(setup, timeout)
+  beforeEach(setupEach, timeout)
+  afterEach(teardownEach, timeout)
+  afterAll(teardownAll, timeout)
+
+  // test(
```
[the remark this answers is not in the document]
fordN: Still commented out code here, please remove if unused.  If you need to keep it around for future work move it to a branch and push it to origin so it stays around.

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -0,0 +1,159 @@
...
+  connectDatabase,
+  createLogger,
+  createMetrics,
+  Logger,
+  Metrics,
+  toAddress,
+} from '@graphprotocol/common-ts'
+import { testNetworkSpecification } from '../../indexer-management/__tests__/util'
+import { Sequelize } from 'sequelize'
+//import { utils } from 'ethers'
```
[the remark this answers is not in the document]
fordN: Still..

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
```diff
@@ -0,0 +1,241 @@
...
+        RETURN NEW;
+    END;
+    $$ LANGUAGE 'plpgsql';
+  `
+  const triggerSQL = `
+    CREATE TRIGGER receipt_update AFTER INSERT OR UPDATE
+    ON scalar_tap_receipts
+    FOR EACH ROW EXECUTE PROCEDURE scalar_tap_receipt_notify();
+  `
+  const indexSQL = `
```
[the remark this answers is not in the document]
fordN: These included in model now?

packages/indexer-agent/src/__tests__/indexer.ts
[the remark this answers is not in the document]
fordN: Doesn't the agent need to know this (contract addresses) in regular operation though too?

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
fordN: Did you add info to the logger? It still looks like a simple copy of the parent logger `const logger = this.logger.child({})`?

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
fordN: I don't follow I'm not suggesting any change in execution just moving the code so it's more readable :)

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
fordN: Still got typo here?

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
fordN: Still using undescriptive name here?

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
fordN: Ok, worth including some actionable suggestion in the comment?

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -89,6 +121,9 @@ export class AllocationReceiptCollector implements ReceiptCollector {
   declare voucherRedemptionBatchThreshold: BigNumber
   declare voucherRedemptionMaxBatchSize: number
   declare protocolNetwork: string
+  declare tapSubgraph: TAPSubgraph
+  declare finalityTime: number
+  declare sequelize: Sequelize
```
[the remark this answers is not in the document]
fordN: Any ideas here? Tried bringing in just QueryInterface instead of entire Sequelize package?

packages/indexer-agent/src/__tests__/indexer.ts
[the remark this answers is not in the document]
carlosvdr: In the real way it uses the npm module to get this values https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-common/src/allocations/query-fees.ts#L14

packages/indexer-agent/src/commands/start.ts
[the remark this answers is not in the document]
carlosvdr: you can see it in here:
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-agent/src/commands/start.ts#L595

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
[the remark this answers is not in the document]
carlosvdr: Like it was proposed it was used using `addIndex()`
here it is used
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts#L72-L75

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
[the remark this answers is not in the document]
carlosvdr: Yes here it is:
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-common/src/query-fees/models.ts#L282-L359

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -367,14 +416,167 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+      )
+      // Lowercase and remove '0x' prefix of addresses to match format in TAP DB Tables
+      const nonRedeemedAllocationIDsTrunc = nonRedeemedAllocationIDAddresses.map(
+        (allocationID) => allocationID.toLowerCase().replace('0x', ''),
+      )
+
+      // Mark RAVs as unredeemed in DB if the TAP subgraph couldn't find the redeem Tx.
+      // To handle a chain reorg that "unredeemed" the RAVs.
+      // WE use sql directly due to a bug in sequelize update:
+      // https://github.com/sequelize/sequelize/issues/7664 (bug been open for 7 years no fix yet or ever)
```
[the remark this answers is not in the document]
carlosvdr: This was done in an older commit in this branch tried it several times and after several tests we decided to move

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
carlosvdr: This was changed in newest commit
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-common/src/allocations/query-fees.ts#L543-L566

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -624,6 +826,126 @@ export class AllocationReceiptCollector implements ReceiptCollector {
     }
   }
 
+  private async submitRAVs(signedRavs: RavWithAllocation[]): Promise<void> {
+    const logger = this.logger.child({
+      function: 'submitRAVs()',
+      signedRavSize: signedRavs.length,
```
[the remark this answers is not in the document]
carlosvdr: BTW this was changed as proposed

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
carlosvdr: want to include a link to create an issue?

packages/indexer-common/src/allocations/query-fees.ts
carlosvdr: This was also changed :
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-common/src/allocations/query-fees.ts#L904-L918

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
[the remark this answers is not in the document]
carlosvdr: This isnt in the models so yes then this can be added in there too
To make it clear all tables in the migrations should be then also in the models right

packages/indexer-agent/src/db/migrations/13-add-scalar-tap-deny-list.ts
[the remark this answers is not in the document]
carlosvdr: Like the above not all were defined in the models so will add them there

packages/indexer-common/src/allocations/__tests__/tap.test.ts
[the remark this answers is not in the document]
carlosvdr: THis has been removed:
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-common/src/allocations/__tests__/tap.test.ts#L1-L18

packages/indexer-common/src/allocations/__tests__/tap.test.ts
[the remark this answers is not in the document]
carlosvdr: This was left out, its addressed in another PR
we planned on adding several more unit tests but didnt wanted to be part of this PR in the end
Addressed here:
https://github.com/graphprotocol/indexer/pull/888

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -54,6 +64,11 @@ interface ReceiptMetrics {
   vouchersRedeemDuration: Histogram<string>
   vouchersBatchRedeemSize: Gauge<never>
   voucherCollectedFees: Gauge<string>
+  successRavRedeems: Counter<string>
```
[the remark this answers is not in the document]
carlosvdr: This was changed as proposed thanks

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
carlosvdr: Yes this has been changed since:
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-common/src/allocations/query-fees.ts#L38

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -110,7 +149,11 @@ export class AllocationReceiptCollector implements ReceiptCollector {
     collector.transactionManager = transactionManager
```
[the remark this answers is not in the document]
carlosvdr: I believe we discussed this through slack, seems it was made this way to prevent direct instantiation of the collector

packages/indexer-agent/src/db/migrations/12-add-scalar-tap-table.ts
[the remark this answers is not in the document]
fordN: Migrations are just for changing data that's already in someone's DB, all table definitions should be in models.

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
carlosvdr: yes it was changed :
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-common/src/allocations/query-fees.ts#L436

packages/indexer-agent/src/db/migrations/13-add-scalar-tap-deny-list.ts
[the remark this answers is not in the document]
fordN: Ok, will of course it was crashing then if there wasn't a model for it....

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
carlosvdr: Yes we understand is not a change in execution, however if we make it more modular its easier to test individually portions of the code when unit testing

packages/indexer-common/src/allocations/__tests__/tap.test.ts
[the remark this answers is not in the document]
fordN: I would really like to no merge commented out code into `main` branch. If you need to keep some code around for later use do so in a branch please.

packages/indexer-agent/src/commands/start.ts
tkornuta-semiotic: @fordN @Jannis no feedback, so resolving

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
carlosvdr: Resolved here:
https://github.com/graphprotocol/indexer/blob/gusinacio/collect-tap-ravs/packages/indexer-common/src/allocations/query-fees.ts#L581

packages/indexer-common/src/allocations/query-fees.ts
carlosvdr: This has been resolved

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
fordN: ok, if that makes it easier to test then that is enough justification to leave as is. 👍

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
carlosvdr: Discussed with Ford, will leave like this.
So resolved.

packages/indexer-agent/src/db/migrations/13-add-scalar-tap-deny-list.ts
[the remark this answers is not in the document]
carlosvdr: Added them to the models now

packages/indexer-common/src/allocations/__tests__/tap.test.ts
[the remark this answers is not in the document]
carlosvdr: Removed the test since it wasnt being used here

packages/indexer-common/src/allocations/query-fees.ts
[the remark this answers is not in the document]
carlosvdr: I included a link to create an issue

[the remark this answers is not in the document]
carlosvdr: https://github.com/graphprotocol/indexer/pull/869/commits/1d869b8bec9b96d876b00ef00874d78aea913f33

packages/indexer-common/src/allocations/query-fees.ts
carlosvdr: Yes this was also causing the same bug as before, I just missed this section too until I hit the error in here too

fordN (approved):
LGTM 👍 😄

PR #871, opened 2024-03-20
chore: use sepolia instead of goerli for tests

saihaj:
locally I had issues running the tests. It turned out we were still using `goerli` which is deprecated now so updating the test RPC to sepolia


fordN (approved):
LGTM 👍

PR #877, opened 2024-04-11
chore: remove outdated Scalar docs

Theodus:
This page describes Scalar in the context of Vector, which is misleading given that Vector is not in use.


fordN (approved):
LGTM 👍

PR #883, opened 2024-04-22
Avoid race conditions when queuing actions (hopefully fixes #879)

### LINKED ISSUE #879: Duplicate actions being added to the actions queue

Jannis:
I noticed that the existing actions that the new actions are compared against to check for duplicates are queried outside the Sequelize transaction that will eventually write the action to the db. We've noticed that duplicates happen when two actions are added to the database in quick succession (<2s). By performing both the existing action queries and the database write within the same transaction, this hopefully won't happen any more.

This hopefully fixes duplicate actions being added to the queue (#879).


fordN (approved):
LGTM 👍

PR #885, opened 2024-04-24
common: issue-878 - wait for block number in confirmation

### LINKED ISSUE #878: Use contract calls when confirming `allocate` action
Currently when confirming an `allocate` action the network subgraph is queried to get the subgraph deployment. If the process of publishing the subgraph to the network and allocating happened very quickly in sequence the subgraph deployment may not be searchable on the network subgraph yet, leading to the following error `SHOULD BE UNREACHABLE: No matching subgraphDeployment (${subgraphDeploymentID.ipfsHash}) found on the network`.  Code [here](https://github.com/graphprotocol/indexer/blob/e65224ed7f3e423bdedab0feeb6dd19d7d8a4855/packages/indexer-common/src/indexer-management/allocations.ts#L998-L1008)

To improve robustness of this step we should update the `matchingRuleExists` function to get this data from the contracts directly (or fall back to the contracts if not available on the network subgraph), so we are not relying on the network subgraph being instantly updated to get this information.  Additionally let's take a minute to audit any other confirmation data and update to this more robust method to avoid similar race conditions.

dwerner:
Resolves https://github.com/graphprotocol/indexer/issues/878 by introducing waiting for the given block to be reached before confirming transactions.

### DIFF
--- a/packages/indexer-common/src/indexer-management/allocations.ts
+++ b/packages/indexer-common/src/indexer-management/allocations.ts
@@ -1,4 +1,5 @@
 import {
+  Eventual,
   formatGRT,
   Logger,
   parseGRT,
@@ -429,6 +430,8 @@ export class AllocationManager {
       )
     }
 
+    await this.waitForBlockNumber(logger, receipt.blockNumber)
+
     const createAllocationEventLogs = this.network.transactionManager.findEvent(
       'AllocationCreated',
       this.network.contracts.staking.interface,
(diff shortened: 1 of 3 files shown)


packages/indexer-common/src/indexer-management/allocations.ts
```diff
@@ -482,6 +485,33 @@ export class AllocationManager {
...
+  }
+
+  async waitForBlockNumberOnEventual(
+    logger: Logger,
+    number: number,
+    sequentialNumberStream: Eventual<number>,
+  ): Promise<void> {
+    for await (const latestSequentialNumber of sequentialNumberStream.values()) {
+      if (latestSequentialNumber >= number) {
+        return
```
fordN, line 505: Are there possible weird edge cases that could lead to this loop running indefinitely? 

If so maybe there is an actionable error message we could present? Or an exit strategy such as a timeout?

dwerner:
choosing to close as this as it hasn't been a big problem

PR #887, opened 2024-05-01
common: 879 duplicate queued actions - composite index

### LINKED ISSUE #879: Duplicate actions being added to the actions queue

dwerner:
Resolves #879

After testing seeing https://github.com/graphprotocol/indexer/pull/883, I set out to write a unit test that would prove if #879 were indeed resolved.

I ended up with a test that would sporadically fail, depending on timing.

My first approach here was to introduce an advisory lock, which I removed in favor of this PR's composite index and unique constraint on deploymentID and source. Now a GQL validation error occurs when the duplicate is about to be inserted.

Looking for feedback on this approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -119,3 +119,4 @@ yalc.lock
 # IDE
 .idea/
 .envrc
+.vscode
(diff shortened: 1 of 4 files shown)


dwerner:
Converting to a draft as I address some test failures.

dwerner:
Closing this PR because it would entail either a constraint for which old data would conflict or a table level lock.

PR #893, opened 2024-05-10
agent,common: lower the default parameters for query fee claiming

cwtaras:


fordN:
Closed in favor of #922

PR #895, opened 2024-05-13
feat: allow early allocation closure

tmigone:


fordN:
Nice, thanks for the PR @tmigone!

fordN (changes requested):
We'll also need to remove [the `createdAtEpoch` filter](https://github.com/graphprotocol/indexer/blob/main/packages/indexer-common/src/operator.ts#L387) from the `closeEligibleAllocations` function.

fordN (approved):

PR #897, opened 2024-05-20
chore: update network subgraph deployment hashes

juanmardefago:
Updated with latest deployment hashes
Can be double checked on their respective endpoints on hosted service


fordN (approved):
LGTM

PR #899, opened 2024-05-24
Allow cost models in single network mode

### LINKED ISSUE #789: Unable to define cost model after upgrading to v0.20.22 indexer SW on arbitrum-one
Hello, 

Since upgraded indexer-agent to v0.20.22 were unable to define cost models. Running single network (arbitrum-one) set up with the INDEXER_AGENT_MULTINETWORK_MODE=false. However both mainnet.tmpl and arbitrum.tmpl are included in the path given to INDEXER_AGENT_NETWORK_SPECIFICATIONS_DIRECTORY.

Logs bellow.

graph indexer -n arbitrum-one cost set model QmbW4ZcQCsJgjdf3sML23XMSP8yZJ9Xi9xVwEgWpbNRC8k /root/tools/costmodels/default.agora
[GraphQL] Can't set cost model: Indexer Agent does not have Ethereum Mainnet network configured.

aasseman:
- `multiNetworks` was erroneously typed as possibly `undefined`, and `undefined` was used as a way to determine if we are in multi or single network mode in `cost-models.ts`.
However, the code doesn't actually allow for `multiNetworks` to be undefined, be it in multi or single network mode.
- The inject DAI setting now defaults to false
- Cost models are allowed only *if in single network mode* and:
  - On Ethereum mainnet
  - On any other network **and** "inject DAI" off

Some of these changes should be inconsequential considering the impending deprecation of Ethereum mainnet, as well as the fact that most indexers stopped using cost models due in part to #789.

Fixes #789

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -43,6 +43,7 @@ const SUGGESTED_SUBGRAPH_MAX_BLOCK_DISTANCE_ON_L2 =
   50 + DEFAULT_SUBGRAPH_MAX_BLOCK_DISTANCE
 const DEFAULT_SUBGRAPH_FRESHNESS_SLEEP_MILLISECONDS = 5_000
 
+// NOTE: This is run only in single-network mode
 export const start = {
   command: 'start',
   describe: 'Start the agent',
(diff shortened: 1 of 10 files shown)


dwerner (approved):
Thanks for the contribution. I had a read through and it all looks good to me.

PR #900, opened 2024-05-29
service: Support setting rate limits and body size limits at startup

fordN:
## Changes
- Added new startup parameters: 
    - **body-size-limit**: Max body size per request (mb), 
    - **info-rate-limit**: Max requests per minute before returning `413` statuses, applies to paths: `/cost`, `/health`, `/operator`,
    - **status-rate-limit**:  Max requests per minute before returning `413` statuses, applies to paths: `/status`, `/network`


dwerner (approved):
lgtm!

packages/indexer-service/src/commands/start.ts
```diff
@@ -185,6 +185,29 @@ export default {
         type: 'string',
         required: false,
       })
+      .option('info-rate-limit', {
+        description:
+          'Max requests per minute before returning 413 statuses, applies to paths: /cost, /health, /operator',
```
dwerner, line 190: 413 status?

PR #904, opened 2024-06-06
Make tapSubgraph to be an optional value

carlosvdr:
Ford found out tapsubgraph is not an optional value so changed a few things to make it work


aasseman (approved):
It's a quick solution, LGTM.

Longer term it'd be nice to have a cleaner solution. In particular it'd be nice to warn the indexer if we see RAVs in the DB but we don't have the right settings to act on them.

fordN:
Nice, thanks for the quick fix! 🙏 

There is still another backwards compatibility issue with the TAP integration though. If the indexer-agent is connected to a network without TAP contracts deployed it fails to startup: runs into an unhandled promise rejection. I'll create an issue to track and we can resolve in a subsequent PR.

fordN (approved):
LGTM! Once it's confirmed working for an indexer without tapSubgraph configured let's get merged in.

packages/indexer-common/src/network.ts
```diff
@@ -143,7 +142,7 @@ export class Network {
     )
 
     const tapSubgraph = new TAPSubgraph(
-      specification.subgraphs.tapSubgraph.url!,
+      specification.subgraphs.tapSubgraph!.url!,
```
fordN, line 145: This assertion is failing. Isn't `specification.subgraphs.tapSubgraph` optional now?

fordN: I caught this one by trying to spin up indexer-agent without a configured `tapSubgraph`. 

Here's the log: 
```
WARN (IndexerAgent/33342): Unhandled promise rejection
    err: {
      "type": "IndexerError",
      "message": "Unhandled promise rejection",
      "stack":
          IndexerError: Unhandled promise rejection
              at indexerError (/Users/ford/thegraph/workspaces/indexer/packages/indexer-common/src/errors.ts:193:10)
              at process.<anonymous> (/Users/ford/thegraph/workspaces/indexer/packages/indexer-agent/src/commands/start.ts:485:24)
              at process.emit (node:events:514:28)
              at process.emit (node:domain:488:12)
              at process.emit.sharedData.processEmitHook.installedValue [as emit] (/Users/ford/.nvm/versions/node/v20.10.0/lib/node_modules/ts-node/node_modules/@cspotcode/source-map-support/source-map-support.js:745:40)
              at emit (node:internal/process/promises:150:20)
              at processPromiseRejections (node:internal/process/promises:284:27)
              at processTicksAndRejections (node:internal/process/task_queues:96:32)
      "code": "IE035",
      "explanation": "https://github.com/graphprotocol/indexer/blob/main/docs/errors.md#ie035",
      "cause": {
        "type": "TypeError",
        "message": "Cannot read properties of undefined (reading 'url')",
        "stack":
            TypeError: Cannot read properties of undefined (reading 'url')
                at Function.create (/Users/ford/thegraph/workspaces/indexer/packages/indexer-common/src/network.ts:145:44)
                at processTicksAndRejections (node:internal/process/task_queues:95:5)
                at async /Users/ford/thegraph/workspaces/indexer/node_modules/p-map/index.js:57:22
      }
    }
```

carlosvdr: Yeah there is logic inside of it , which will take care in case its not defined, but it is confusing to read you are right.
Added just a flag instead in this part

fordN: It didn't get to actually execute the function since this is erroring on an assertion.
 
 Should we just remove that assertion? Could use `?` rather than `!`?

fordN: Or we can just skip it altogether if config value isn't present. Something like [this](https://github.com/graphprotocol/indexer/pull/904/commits/897e1ca48812f7b4df6ff012a3d5b7fb6acfef29)?

PR #914, opened 2024-06-17
indexer-common: Avoid using createdAt field for pagination

fordN:
## Changes
- Update several `NetworkMonitor` functions to paginate by `id` instead of by `createdAt`. Functions updated: subgraphs(), subgraphDeployments() and disputableAllocations(). 

## Background
On L2 networks, like arbitrum-one, `createdAt` is not necessarily unique, so paginating based on `createdAt` can lead to missing items.

### DIFF
--- a/packages/indexer-common/src/indexer-management/monitor.ts
+++ b/packages/indexer-common/src/indexer-management/monitor.ts
@@ -336,7 +336,7 @@ export class NetworkMonitor {
     }
     let subgraphs: Subgraph[] = []
     const queryProgress = {
-      lastCreatedAt: 0,
+      lastId: '',
       first: 20,
       fetched: 0,
       exhausted: false,
@@ -352,10 +352,10 @@ export class NetworkMonitor {
       try {
         const result = await this.networkSubgraph.checkedQuery(
           gql`
-            query subgraphs($first: Int!, $lastCreatedAt: Int!, $subgraphs: [String!]!) {
+            query subgraphs($first: Int!, $lastId: String!, $subgraphs: [String!]!) {
               subgraphs(
-                where: { id_gt: $lastCreatedAt, id_in: $subgraphs }
-                orderBy: createdAt
+                where: { id_gt: $lastId, id_in: $subgraphs }
+                orderBy: id
                 orderDirection: asc
                 first: $first
               ) {
@@ -374,7 +374,7 @@ export class NetworkMonitor {
           `,
           {
             first: queryProgress.first,
-            lastCreatedAt: queryProgress.lastCreatedAt,
+            lastId: queryProgress.lastId,
             subgraphs: ids,
           },
         )
@@ -410,7 +410,7 @@ export class NetworkMonitor {
 
         queryProgress.exhausted = results.length < queryProgress.first
         queryProgress.fetched += results.length
-        queryProgress.lastCreatedAt = results[results.length - 1].createdAt
+        queryProgress.lastId = results[results.length - 1].id
 
         subgraphs = subgraphs.concat(results)
       } catch (error) {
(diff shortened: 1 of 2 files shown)


dwerner (approved):
lgtm! thanks

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -336,7 +336,7 @@ export class NetworkMonitor {
     }
     let subgraphs: Subgraph[] = []
     const queryProgress = {
-      lastCreatedAt: 0,
+      lastId: 0,
```
aasseman, line 339: The network subgraph's schema is using type `ID` for ids, which is an alias for `string`.
(see https://semiotic-ai.slack.com/archives/C032EDF093K/p1713996760318369?thread_ts=1713983989.035979&cid=C032EDF093K)

```suggestion
      lastId: '',
```

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -352,10 +352,10 @@ export class NetworkMonitor {
       try {
         const result = await this.networkSubgraph.checkedQuery(
           gql`
-            query subgraphs($first: Int!, $lastCreatedAt: Int!, $subgraphs: [String!]!) {
+            query subgraphs($first: Int!, $lastId: Int!, $subgraphs: [String!]!) {
```
aasseman, line 355:
```suggestion
            query subgraphs($first: Int!, $lastId: ID!, $subgraphs: [String!]!) {
```

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -596,10 +596,10 @@ export class NetworkMonitor {
       try {
         const result = await this.networkSubgraph.checkedQuery(
           gql`
-            query subgraphDeployments($first: Int!, $lastCreatedAt: Int!) {
+            query subgraphDeployments($first: Int!, $lastId: Int!) {
```
aasseman, line 599:
```suggestion
            query subgraphDeployments($first: Int!, $lastId: ID!) {
```

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -1141,15 +1140,15 @@ Please submit an issue at https://github.com/graphprotocol/block-oracle/issues/n
             ) {
```
aasseman, line 1140: This is missing the variable declaration for `$lastId`. 
`graph-node` may let it pass, but I think that The Guild is/was working on making graph-node stricter wrt. the GraphQL spec, which could cause issues in the future.

aasseman (changes requested):
Just nitpicking on adhering to the subgraph's schema, though graph-node may currently allow these things.

Also, somewhat unrelated, I found a few places in the code where pagination isn't implemented at all. In particular that could cause issues if somehow an indexer managed to open more than 1000 allocations. We know that some indexers are already indexing more than 1000 subgraphs at a time, so it's not completely impossible.

dwerner:
Removing my approval as I've made changes now.

aasseman (changes requested):
One of the comments was not addressed. GitHub won't let me explicitly select the piece of code....

So here's the issue:
`packages/indexer-common/src/indexer-management/monitor.ts:1139` `$createdAt: Int!` should be `$lastId: ID!`

dwerner:
> One of the comments was not addressed. GitHub won't let me explicitly select the piece of code....
> 
> So here's the issue: `packages/indexer-common/src/indexer-management/monitor.ts:1139` `$createdAt: Int!` should be `$lastId: ID!`

Thanks for catching this.

aasseman (approved):

dwerner (approved):
Looks good to me.

PR #918, opened 2024-06-20
Add missing addresses and resolve table formatting issues in network docs

alex-pakalniskis:


fordN:
Nice, thanks for PR @alex-pakalniskis. 🙌 
Can we cleanup commit history before merging? We should update commit messages to match our format and squash the last 2 into 1.

So the commit history may look like:
- `docs: Add DataEdge and Block Oracle to Arb-One docs`
- `docs: Fix markdown formatting of tables`

PR #867, opened 2024-03-04
indexer-agent: Leave subgraph deployment assignment decisions to graph-node

fordN:
[Hands off the assignments](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExbTd2ZHhyenNxd29sYjE3aWdxNWFxbDVrb3R5azF5bXBybWYyMDd2bSZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/jtKYsTf4BtrzzaMfMT/giphy.gif)

Resolve #853 

### Changes
- Use recently added subgraph_pause method to explicitly pause subgraph deployments rather than virtually pausing them by assigning to a non-existent index node
- Remove index-node-ids startup parameter from indexer-agent
- Never send an index-node-id param with subgraph_deploy requests, the receiving graph-node will make the index-node-id decision based on the graph-node.toml config file

PR #920, opened 2024-06-22
Indexer-common: Fix graph-node deployment rules; remove reassign command on initial deployment

datanexus-vincent:
Wanting to use the graph-node deployment rules (as shown [here](https://github.com/graphprotocol/graph-node/blob/master/docs/config.md#controlling-deployment)), I followed the logic path for the `subgraph_deploy` graph-node RPC method. I found that it already selects an initial deployment based on the deployment rules, ignoring any deployment target you pass with the RPC call. So I took a look at the indexer agent program, and I found that after it calls the `subgraph_deploy` RPC method, it calls the `subgraph_reassign` method, which doesn't ignore the target node you pass with the call.

Removing the redeploy call leaves the graph-node to decide the target deployment based on the deployment rules defined in your graph node config, as it should be.

### DIFF
--- a/packages/indexer-common/src/graph-node.ts
+++ b/packages/indexer-common/src/graph-node.ts
@@ -383,7 +383,6 @@ export class GraphNode {
           })[0].id
       await this.create(name)
       await this.deploy(name, deployment, targetNode)
-      await this.reassign(deployment, targetNode)
     } catch (error) {
       if (!(error instanceof IndexerError)) {
         const errorCode = IndexerErrorCode.IE020


datanexus-vincent:
I just saw https://github.com/graphprotocol/indexer/pull/867, so I'm going to close this PR since that is a much more thorough fix. I will post this as a patch in discord though, since this is a quick fix that will work nicely until the more thorough fix is merged.

For posterity, this is the graph-node logic I tracked down that I knew it would work:
https://github.com/graphprotocol/graph-node/blob/d9b37b30f38f4647fa718fbd32de6322a181032f/store/postgres/src/subgraph_store.rs#L533
There you can see it ignores the node_id it passes through when deploying a subgraph. And I traced that from the subgraph_deploy RPC call handler here:
https://github.com/graphprotocol/graph-node/blob/d9b37b30f38f4647fa718fbd32de6322a181032f/server/json-rpc/src/lib.rs#L122
And there's also a mention here that it shouldn't even be accepting a node_id as a parameter.
https://github.com/graphprotocol/graph-node/blob/d9b37b30f38f4647fa718fbd32de6322a181032f/store/postgres/src/subgraph_store.rs#L1282

My simple patch:
```
diff --git a/packages/indexer-common/src/graph-node.ts b/packages/indexer-common/src/graph-node.ts
index 62cc6fa..9c10a8c 100644
--- a/packages/indexer-common/src/graph-node.ts
+++ b/packages/indexer-common/src/graph-node.ts
@@ -383,7 +383,6 @@ export class GraphNode {
           })[0].id
       await this.create(name)
       await this.deploy(name, deployment, targetNode)
-      await this.reassign(deployment, targetNode)
     } catch (error) {
       if (!(error instanceof IndexerError)) {
         const errorCode = IndexerErrorCode.IE020
```

PR #921, opened 2024-06-22
feat: Deployment tags to replace `indexer-agent` prefix in deployment names

datanexus-vincent:
Indexers need an easy way to deploy to different shards and graph-nodes. This is already possible with graph-node alone via deployment rules and deployment name regex matching, but the indexer-agent doesn't give any way to control the name that a deployment is given. With this PR, I've introduced a `tag` column for indexing rules, whose contents will replace the `indexer-agent` prefix for deployment names upon first deployment.

With a combination of a deployment rule like this:
```
[[deployment.rule]]
match = { name = "(vip|important)/.*", network = "arbitrum-one" }
shard = "arbitrum_vip"
indexers = [ "graph_node_vip_0", "graph_node_vip_1" ]
```
and a `tag` set like this:
```
graph indexer rules set QmHash decisionBasis offchain tag vip
```
You can now deploy VIP arbitrum subgraphs to your arbitrum VIP database directly.

Feel free to request or make changes. I'll do my best to satisfy any requirements so we can get this merged.

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -738,11 +738,12 @@ export class Agent {
         switch (this.deploymentManagement) {
           case DeploymentManagementMode.AUTO:
             try {
+              const resolvedDeploymentTags = await deploymentTags.value()
               await this.reconcileDeployments(
                 activeDeployments,
                 targetDeployments,
                 eligibleAllocations,
-                deploymentTags,
+                resolvedDeploymentTags,
               )
             } catch (err) {
               logger.warn(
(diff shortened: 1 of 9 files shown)


fordN:
Love this idea, @datanexus-vincent! 🙌 

We'll also need some way of representing this in the actions queue so that deployments created/deployed through that flow can have tags applied too.  I'll take a closer look at how to make that addition and see if we can get this in next week.

suntzu93:
@fordN Is there any plan to implement this yet?

trader-payne:
Gentle nudge from me too 👀

PR #922, opened 2024-06-24
agent,common: lower the default parameters for query fee claiming

chriswessels:
Repushing @cwtaras PR #893 as a maintainer to check if CI actions pass


dwerner (approved):
lgtm

PR #930, opened 2024-06-26
feat: add history for cost models and set view table for latest models

carlosvdr:
Reason for this PR. 

tl;dr: needed for TAP cost model checks

We currently have an issue where when a cost model just got updated  the indexer-service takes some time to update but in the meantime it errors out due to the model not matching, so we need to have a history in the db so that we can accept both, the new and the old one for a delimited amount of time (eg. ~2min)
This PR includes a breaking change:

- CostModels table no longer exists
- cost_models table is created, this will contain a history of all past cost models
- CostModels table will keep working as a view from cost_models this way other programs that read into CostModels will keep working as they are, this view will contain only the latest model of each deployment thus not breaking anything else

### DIFF
--- a/packages/indexer-agent/src/db/migrations/14-cost-models-history.ts
+++ b/packages/indexer-agent/src/db/migrations/14-cost-models-history.ts
@@ -1,0 +1,150 @@
+import { Logger } from '@graphprotocol/common-ts'
+import { utils } from 'ethers'
+import { QueryInterface, DataTypes } from 'sequelize'
+
+interface MigrationContext {
+  queryInterface: QueryInterface
+  logger: Logger
+}
+
+interface Context {
+  context: MigrationContext
+}
+export const COST_MODEL_GLOBAL = 'global'
+export async function up({ context }: Context): Promise<void> {
+  const { queryInterface, logger } = context
+
+  const tables = await queryInterface.showAllTables()
+  logger.debug(`Checking if CostModelsHistory table exists`, { tables })
+
+  // CostModelsHistory: this table will store the history of cost models
+  // this is necessary since there could be a mismtach between the old table and the info the gateway has
+  // causing a failed request. Solution is to have a history and allow the "old" model for a limited time frame 2-3 mins
+  // For indexer-service is also helpful to have the history of the cost models since we want to obtain the minimum cost per receipt
+  // this will help since the gateway could send an old model and get blocked so we need the indexer to accept one of the 2 latest models
+  // in the past 30 seconds since the gateway updates its model every 30 seconds
+
+  if (tables.includes('CostModelsHistory')) {
+    logger.debug(`CostModelsHistory already exist, migration not necessary`)
+  } else {
+    logger.info(`Create CostModelsHistory`)
+    await queryInterface.createTable('CostModelsHistory', {
+      id: {
+        type: DataTypes.BIGINT,
+        primaryKey: true,
+        autoIncrement: true,
+        unique: true,
+      },
+      deployment: {
+        type: DataTypes.STRING,
+        allowNull: false,
+        validate: {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          isDeploymentID: (value: any) => {
+            if (typeof value !== 'string') {
+              throw new Error('Deployment ID must be a string')
+            }
+            // "0x..." and "global" is ok
+            if (utils.isHexString(value, 32) || value === COST_MODEL_GLOBAL) {
+              return
+            }
+
+            throw new Error(
+              `Deployment ID must be a valid subgraph deployment ID or "global"`,
+            )
+          },
+        },
+      },
+      model: {
+        type: DataTypes.TEXT,
+        allowNull: true,
+      },
+      variables: {
+        type: DataTypes.JSONB,
+        allowNull: true,
+      },
+      createdAt: {
+        type: DataTypes.DATE,
+        allowNull: false,
+      },
+      updatedAt: {
+        type: DataTypes.DATE,
+        allowNull: false,
+      },
+    })
+    if (tables.includes('CostModels')) {
+      logger.debug(`Copying data from CostModels into CostModelsHistory`)
+      const copyTableSQL = `
+          INSERT INTO "CostModelsHistory"
+          SELECT * FROM "CostModels";
+      `
+      await queryInterface.sequelize.query(copyTableSQL)
+      logger.info(`Drop table "CostModels"`)
+      await queryInterface.dropTable('CostModels', { cascade: true })
+    }
+    // To avoid creating a breaking change for the indexer-agent or indexer-service we create a view table
+    // Since now other systems can keep the same query towards "CostModels" and not need to change anything
+
+    logger.info(
+      `Creating a view for CostModelsHistory to substitute "CostModels" table`,
+    )
+    const viewSQL = `
+      CREATE VIEW "CostModels" AS SELECT id,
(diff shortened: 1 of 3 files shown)


aasseman (changes requested):

packages/indexer-agent/src/db/migrations/14-cost-models-history.ts
```diff
@@ -17,6 +17,10 @@ export async function up({ context }: Context): Promise<void> {
   const tables = await queryInterface.showAllTables()
   logger.debug(`Checking if CostModelsHistory table exists`, { tables })
 
+  // CostModelsHistory: this table will store the history of cost models
+  // this is necessary since there could be a mismtach between the old table and the info the gateway has
+  // causing a failed request. Solution is to have a history and allow the "old" model for a limited time frame 2-3 mins
```
aasseman, line 22: You also need to give the context about tap-agent checking the receipt values

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -80,14 +80,21 @@ export default {
...
+       "updatedAt"
+      FROM "CostModelsHistory" t1
+      JOIN
+      (
+          SELECT MAX(id)
+          FROM "CostModelsHistory"
+          ${deployments ? 'WHERE deployment IN (:deployments)' : ''}
+          GROUP BY deployment
+      ) t2
+        ON t1.id = t2.MAX;
```
aasseman, line 97: Can you also use this query in the view?

aasseman (changes requested):

packages/indexer-agent/src/db/migrations/14-cost-models-history.ts
```diff
@@ -0,0 +1,147 @@
...
+      updatedAt: {
+        type: DataTypes.DATE,
+        allowNull: false,
+      },
+    })
+    if (tables.includes('CostModels')) {
+      logger.debug(`Copying data from CostModels into CostModelsHistory`)
+      const copyTableSQL = `
+          INSERT INTO "CostModelsHistory" (id, deployment, model, variables, "createdAt", "updatedAt") 
+          SELECT id, deployment, model, variables, "createdAt", "updatedAt" FROM "CostModels";
```
aasseman, line 79: You shouldn't need to copy the ID, just let PG create new ones

packages/indexer-agent/src/db/migrations/14-cost-models-history.ts
```diff
@@ -0,0 +1,147 @@
...
+       "updatedAt"
+      FROM "CostModelsHistory" t1
+      JOIN
+      (
+          SELECT MAX(id)
+          FROM "CostModelsHistory"
+          GROUP BY deployment
+      ) t2
+        ON t1.id = t2.MAX;
+      ;
```
aasseman, line 106: Extra `;`

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -73,14 +75,38 @@ export default {
...
+      ;
```
aasseman, line 98: extra `;`

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -121,22 +146,17 @@ export default {
     } catch (err) {
       throw new Error(`Invalid cost model or variables: ${err.message}`)
     }
-    const [model] = await models.CostModel.findOrBuild({
+    const [oldModel] = await models.CostModel.findOrBuild({
```
aasseman, line 149: Probably better not to build. Just try to find the latest model, and if there isn't one don't try to copy values over.

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -121,22 +146,17 @@ export default {
...
       where: { deployment: update.deployment },
+      order: [['createdAt', 'DESC']],
```
aasseman, line 151: Order by ID instead. Last ID will always be the freshest one.

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -121,22 +146,17 @@ export default {
...
-    // logger.info('Merged models', { now: model })
-    model.deployment = update.deployment || model.deployment
-    model.model = update.model || model.model
-
-    // Update the model variables (fall back to current value if unchanged)
-    let variables = update.variables || model.variables
+    const model = models.CostModel.build()
 
+    model.deployment = update.deployment || oldModel.deployment
+    model.model = update.model || oldModel.model
```
aasseman, line 156: I think it'd be cleaner to create the new entry in one go with by giving all the values in `build`, instead of creating then updating the row column by column.
Example from the [sequelize docs](https://sequelize.org/docs/v6/core-concepts/getters-setters-virtuals/#setters):

```js
const user = User.build({
  username: 'someone',
  password: 'NotSo§tr0ngP4$SW0RD!',
});
console.log(user.password); // '7cfc84b8ea898bb72462e78b4643cfccd77e9f05678ec2ce78754147ba947acc'
console.log(user.getDataValue('password')); // '7cfc84b8ea898bb72462e78b4643cfccd77e9f05678ec2ce78754147ba947acc'
```

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -73,14 +75,38 @@ export default {
...
+    const costModels = await sequelize.query(query, {
+      replacements: { deployments: deployments ? deployments : [] },
+      mapToModel: true,
+      model: models.CostModel,
     })
     const definedDeployments = new Set(costModels.map((model) => model.deployment))
     const undefinedDeployments = deployments?.filter((d) => !definedDeployments.has(d))
     const globalModel = await models.CostModel.findOne({
       where: { deployment: COST_MODEL_GLOBAL },
+      order: [['createdAt', 'DESC']],
```
aasseman, line 109: Order by ID instead

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -52,13 +52,15 @@ export default {
...
       where: { deployment },
+      order: [['createdAt', 'DESC']],
     })
     if (model) {
       return model.toGraphQL()
     }
 
     const globalModel = await models.CostModel.findOne({
       where: { deployment: COST_MODEL_GLOBAL },
+      order: [['createdAt', 'DESC']],
```
aasseman, line 63: Order by ID instead

packages/indexer-common/src/indexer-management/resolvers/cost-models.ts
```diff
@@ -52,13 +52,15 @@ export default {
   ): Promise<object | null> => {
     const model = await models.CostModel.findOne({
       where: { deployment },
+      order: [['createdAt', 'DESC']],
```
aasseman, line 55: Order by ID instead

aasseman (changes requested):

aasseman (commented):
LGTM, please squash everything into a single commit so we can get the maintainers to review

aasseman (approved):

dwerner:
looks good but let's get this in after we tag next release

dwerner (approved):

PR #941, opened 2024-07-05
fix: metrics called before verifying if there are any receipts

carlosvdr:
A metric that tracks number of receipts is being called before checking if any receipt is ready to be collected.

- This could lead in misleading or unnecessary data being generated.
- It could cause type mismatch error between `Address` (expected) and `undefined`

Thus a refactor for this call after the check


aasseman:
Not confirmed to be the source of observed errors in testnet testing, closing for now.

Issue #946, opened 2024-07-10
Figure out Issue regarding sequelize


carlosvdr:
Vincent found this error in the logs
```
{"level":50,"time":1720058625559,"pid":100623,"hostname":"testnet-indexer-agent","name":"IndexerAgent","component":"Network","indexer":"0x86868BB556DeDcd0768854A5aF9e62f2e9129D46","protocolNetwork":"eip155:421614","operator":"0xbC2c5C4Ea1db64f34d61a40cf325D062CFd91f42","component":"AllocationReceiptCollector","err":{"type":"TypeError","message":"Cannot read properties of undefined (reading 'sequelize')","stack":"TypeError: Cannot read properties of undefined (reading 'sequelize')\n    at AllocationReceiptCollector.pendingRAVs (/home/ubuntu/indexer/packages/indexer-common/dist/allocations/query-fees.js:373:39)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)\n    at async tryMap.onError.logger.error.err.err (/home/ubuntu/indexer/packages/indexer-common/dist/allocations/query-fees.js:266:26)"},"msg":"Failed to query pending RAVs"}
```
which makes reference to this part of the [code](https://github.com/graphprotocol/indexer/blob/main/packages/indexer-common/src/allocations/query-fees.ts#L617C7-L622C55)
```
let query = `
        UPDATE scalar_tap_ravs
        SET redeemed_at = NULL
        WHERE allocation_id IN ('${nonRedeemedAllocationIDsTrunc.join("', '")}')
      `
      await this.queryInterface.sequelize.query(query)
```

PR #948, opened 2024-07-10
fix: remove queryInterface reference

carlosvdr:
Regarding issue #946 
Instead of passing along a reference better to just use the sequelize reference inside the model object

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -593,14 +593,7 @@ export async function run(
   const networks: Network[] = await pMap(
     networkSpecifications,
     async (spec: NetworkSpecification) =>
-      Network.create(
-        logger,
-        spec,
-        queryFeeModels,
-        graphNode,
-        metrics,
-        sequelize.getQueryInterface(),
-      ),
+      Network.create(logger, spec, queryFeeModels, graphNode, metrics),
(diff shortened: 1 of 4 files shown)


fordN (approved):
Nice, LGTM 👍

fordN:
Not clear if it resolves [the referenced issue](https://github.com/graphprotocol/indexer/issues/946) though.  Were you able to reproduce that?

carlosvdr:
Wasn't able to reproduce it, we did some digging about it and it seems it was some PG connection related issue which is out of our hands.
Vincent who found this, just restarted and the problem was solved.
But while doing that I realized there is a better way of doing the raw queries.
What im proposing is instead of creating  queryInterface from the sequelize object
To create it from the model
Before i would create a reference of sequelize's query called **queryInterface** which allows me to handle raw queries
I would pass along this object all the way to the **AllocationReceiptCollector**
So the change would be instead of passing **queryInterface** , i would directly query through sequelize reference which is contained inside the model thus getting through the process of passing along **queryInterace** to just raw query like this
`this.models.receiptAggregateVouchers.sequelize?.query(query)`
which would simplify the code
I believe doing this diminishes the possibility of that problem since we are not passing along any more extra references but using what we already had from before (coming from the model), which we know is pretty stable

PR #952, opened 2024-07-12
fix: use tapSubgraph to ignore connection with tapContracts

carlosvdr:


packages/indexer-common/src/network.ts
```diff
@@ -244,17 +244,18 @@ export class Network {
...
-    let tapContracts: any
-    try {
-      tapContracts = await connectTapContracts(
-        wallet,
-        networkIdentifier.chainId,
-        specification.tapAddressBook,
-      )
-    } catch (err) {
-      logger.error(`Failed to connect to tap contract bindings:`, { err })
+    let tapContracts: any = undefined
```
fordN, line 247: We do know what type this would be if not undefined though, right? 

So we can do:
`let tapContracts: TapContracts | undefined = undefined`

packages/indexer-common/src/network.ts
```diff
@@ -244,17 +244,18 @@ export class Network {
...
+    if (tapSubgraph) {
+      try {
+        tapContracts = await connectTapContracts(
+          wallet,
+          networkIdentifier.chainId,
+          specification.tapAddressBook,
+        )
+      } catch (err) {
+        logger.error(`Failed to connect to tap contract bindings:`, { err })
```
fordN, line 256: If there is a failure to connect do we want to only log and continue? Or should we exit out at that point?

carlosvdr: I guess you are right, in that point because if we do send a subgraph url we definitely desire TAP to work, and if contracts don't work what is the point

aasseman: Yup, I'd prefer to see a "fail fast" here

fordN:
Please use commit format we've been trying to stick to:
```
component name: Description

Further description if needed
```


So first commit message would be something like:

```
common: Use tapSubgraph config to control TAP/RAV feature

If a user does not config tapSubgraph, do not connect to TAP contracts or
execute RAV processing functions
```

PR #953, opened 2024-07-12
Network Docs: Remove reference to deprecated ethereum-network param, update gateway params

fordN:
## Changes
- Remove reference to --ethereum-network param (no longer used)
- Update gateway endpoint params, now use a single url and use that to build the paths (such as /collect-receipts)


dwerner (approved):
LGTM :+1:

PR #955, opened 2024-07-12
common, docs: Update subgraph maxBlockDistance default values

fordN:
## Background
The default values were far too strict, leading to the agent operations halting.

## Changes
- Update default `maxBlockDistance`:  0 --> 1000.
- Update default `freshnessSleepMilliseconds`: 5,000 --> 10,000
- Update subgraph-freshness feature docs to reflect change in defaults.


dwerner (approved):
:+1:

Issue #972, opened 2024-07-29
Bug: Seg fault starting Node processes


fordN:
Observed by several indexers when trying to run `indexer-agent` or `indexer-service`.

The error appears to be coming from `neon` libraries.

dwerner:
```
(No debugging symbols found in /usr/bin/env)
(gdb) run
Starting program: /usr/bin/env node /home/pops/git/indexer/packages/indexer-service/bin/graph-indexer-service
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
process 1468493 is executing new program: /usr/bin/node
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff7a47640 (LWP 1468514)]
[New Thread 0x7ffff7246640 (LWP 1468515)]
[New Thread 0x7ffff6a45640 (LWP 1468516)]
[New Thread 0x7ffff6244640 (LWP 1468517)]
[New Thread 0x7ffff5a43640 (LWP 1468518)]
[New Thread 0x7ffff5242640 (LWP 1468519)]

Thread 1 "node" received signal SIGSEGV, Segmentation fault.
0x00007ffff4c3dd83 in neon::object::class::ClassMap::get () from /home/pops/git/indexer/packages/indexer-native/binary/index.node
(gdb) bt 0x00007ffff4c3dd83
#0  0x00007ffff4c3dd83 in neon::object::class::ClassMap::get () from /home/pops/git/indexer/packages/indexer-native/binary/index.node
#1  0x00007ffff4c0fdf8 in neon::object::class::Class::constructor () from /home/pops/git/indexer/packages/indexer-native/binary/index.node
#2  0x00007ffff4c10326 in indexer_native::__LOAD_NEON_MODULE::__init_neon_module () from /home/pops/git/indexer/packages/indexer-native/binary/index.node
#3  0x00007ffff4c3ec4d in neon::context::internal::initialize_module () from /home/pops/git/indexer/packages/indexer-native/binary/index.node
#4  0x0000000000ad59e7 in std::_Function_handler<bool (node::binding::DLib*), node::binding::DLOpen(v8::FunctionCallbackInfo<v8::Value> const&)::{lambda(node::binding::DLib*)#1}>::_M_invoke(std::_Any_data const&, node::binding::DLib*&&) ()
[...]

PR #975, opened 2024-07-30
fix(indexer-native): neon 1.0.0, napi-6

dwerner:
In working to resolve https://github.com/graphprotocol/indexer/issues/972, I pulled in some changes from https://github.com/graphprotocol/indexer/pull/749, while retaining our prebuilt pipeline.

### DIFF
--- a/packages/indexer-native/native/Cargo.toml
+++ b/packages/indexer-native/native/Cargo.toml
@@ -1,27 +1,21 @@
 [package]
 name = "indexer-native"
-version = "0.1.0"
-authors = ["Zac Burns <That3Percent@gmail.com>"]
+version = "0.2.0"
+authors = [
+  "Zac Burns <That3Percent@gmail.com>",
+  "Theo Butler <theodusbutler@gmail.com>",
+]
 license = "MIT"
-build = "build.rs"
 edition = "2018"
-exclude = ["artifacts.json", "index.node"]
 
 [lib]
-name = "indexer_native"
 crate-type = ["cdylib"]
 
-[build-dependencies]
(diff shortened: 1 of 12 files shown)


packages/indexer-native/package.json
```diff
@@ -59,12 +41,5 @@
     "jest": "<30.0.0-0",
     "prettier": "3.0.3"
   },
-  "binary": {
```
fordN: We may want to keep this in since it is used to configure our binary build workflow.

packages/indexer-native/native/Cargo.toml
```diff
@@ -6,9 +6,10 @@ authors = [
   "Theo Butler <theodusbutler@gmail.com>",
 ]
 license = "MIT"
-edition = "2018"
+edition = "2021"
```
fordN, line 9: What's the importance of 2021 here?

packages/indexer-native/package.json
```diff
@@ -16,13 +16,31 @@
...
+    "darwin",
+    "linux"
+  ],
+  "cpu": [
+    "x64",
+    "arm",
+    "arm64"
+  ],
+  "engines": {
+    "node": ">=12.22.0"
```
fordN, line 29: worth bumping? This setting can be overly strict sometimes though

fordN (approved):
Looking good!

packages/indexer-native/native/Cargo.toml
dwerner: It's not important if we're retiring this... but keeping this dialect up to date is nice if we weren't.

PR #978, opened 2024-08-02
agent: defend against graph-node bug #5550 (null paused)

dwerner:
Resolves an issue where graph-node returns an error when querying deployment assignment statuses and the `paused` value is null. 

https://github.com/graphprotocol/graph-node/issues/5550


fordN (approved):
LGTM 👍

PR #979, opened 2024-08-05
docs: Add 2 chains to feature matrix: fuse & blast

p-diogo:
Adds full protocol support for ~`moonbeam`,~ `fuse`, and `blast` data sources. 

The Council will be reviewing the CIP reports for these chains this Thursday. 

Next steps:
- [x]  Create a [GGP](https://snapshot.org/#/council.graphprotocol.eth) referencing this PR (Council approving full protocol support for these 3 chains)
- [x]  Wait for the Council to vote
- [x]  Add GGP to the feature matrix 
- [x]   Merge PR

EDIT: removed `moonbeam` from the list.

### DIFF
--- a/docs/feature-support-matrix.md
+++ b/docs/feature-support-matrix.md
@@ -27,7 +27,6 @@ The matrix below reflects the canonical Council-ratified version. As outlined in
 | eip155:534352              | scroll        | Yes         | Yes          | Yes               | Yes                  | Yes              |
 | eip155:59144               | linea         | Yes         | Yes          | Yes               | Yes                  | Yes              |
 | eip155:56                  | bsc           | Yes         | Yes          | Yes               | Yes                  | Yes              |
-| eip155:1284                | moonbeam      | Yes         | Yes          | Yes               | Yes                  | Yes              |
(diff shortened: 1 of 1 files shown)


abourget:
As discussed recently with @mangas .. there's an issue with enabling Substreams-powered Subgraphs for names like "moonbeam" is tricky because those chains do not have Extended traces instrumentation..

This creates a problem of data provenance (the source backing a subgraph) needs to be holistically defined, and if we go from Light blocks to Extended blocks, all under the "moonbeam" network name, things will break determinism.

The `graph-node` is able to abstract from one to the other (offers triggers from both sources), but Substreams get fed the big raw blocks, and those need to be deterministic.

I think it's not safe to enable SpS without defining the exact meaning of `moonbeam`, and then be ready for, in the future, support `moonbeam+extended` (and consider `moonbeam` to _always_ be light blocks).

PedroMD:
As we're still debating how to best handle different block types (base or extended, related to @abourget message above), I'm removing `moonbeam` so we can proceed with the other non-contentious chains.

PR #980, opened 2024-08-05
common: add moonbeam, fuse, blast to chain id lookups 

PedroMD:
## Changes
- Add `moonbeam`, `fuse`, and `blast`  to chain id lookups. (in preparation for https://github.com/graphprotocol/indexer/pull/979)

### DIFF
--- a/packages/indexer-common/src/indexer-management/types.ts
+++ b/packages/indexer-common/src/indexer-management/types.ts
@@ -184,6 +184,9 @@ export const Caip2ByChainAlias: { [key: string]: string } = {
   linea: 'eip155:59144',
   scroll: 'eip155:534352',
   base: 'eip155:8453',
+  moonbeam: 'eip155:1284',
+  fuse: 'eip155:122',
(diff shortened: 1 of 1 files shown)


packages/indexer-common/src/indexer-management/types.ts
```diff
@@ -184,6 +184,9 @@ export const Caip2ByChainAlias: { [key: string]: string } = {
   linea: 'eip155:59144',
   scroll: 'eip155:534352',
   base: 'eip155:8453',
+  moonbeam: 'eip155:1284',
+  fuse: 'eip155:122',
+  blast: 'eip155:81457',
 }
 
 export const Caip2ByChainId: { [key: number]: string } = {
```
fordN, line 192: We'll also need to add corresponding entries to the `Caip2ByChainID` mapping.

PedroMD: oh, I missed it 🤦 fixing now

dwerner (approved):

PR #981, opened 2024-08-05
common: Improve collision resistance of allocation ids

fordN:
Allow user to allocate multiple times to a deployment within the same epoch.

## Change
- Pass ids of recently closed allocations into `uniqueAllocationID()` to avoid collisions with allocations open earlier in the same epoch for the same deployment.


dwerner (approved):
lgtm :+1:

PR #983, opened 2024-08-08
fix: Remove failed subgraph check

datanexus-vincent:
Failed subgraphs still need to be allocated to in order to support historical data queries until the subgraph can be fixed and upgraded. Please remove this check.


fordN:
Closing, #984 resolves this

PR #986, opened 2024-08-16
indexer-common: fix paused mechanism

dwerner:
Corrects the query responsible for checking the graph-node indexingStatus.


fordN (approved):
👍

PR #987, opened 2024-08-16
feat: Add new column to invalid_receipts table

carlosvdr:
Added new column to keep track of why each invalid receipt failed, for easier future debbuging

### DIFF
--- a/packages/indexer-agent/src/db/migrations/16-modify-invalid-receipts-table.ts
+++ b/packages/indexer-agent/src/db/migrations/16-modify-invalid-receipts-table.ts
@@ -1,0 +1,34 @@
+import { Logger } from '@graphprotocol/common-ts'
+import { QueryInterface, DataTypes } from 'sequelize'
+
+interface MigrationContext {
+  queryInterface: QueryInterface
+  logger: Logger
+}
+
+interface Context {
+  context: MigrationContext
+}
+
+export async function up({ context }: Context): Promise<void> {
+  const { queryInterface, logger } = context
+
+  const tables = await queryInterface.showAllTables()
+  logger.debug(
+    `Modifying tables scalar_tap_receipts_invalid to add extra column`,
+  )
+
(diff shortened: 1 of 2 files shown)


packages/indexer-agent/src/db/migrations/16-modify-invalid-receipts-table.ts
```diff
@@ -0,0 +1,51 @@
...
+  )
+  await queryInterface.removeIndex(
+    'scalar_tap_receipts',
+    'scalar_tap_receipts_allocation_id_idx',
+  )
+  await queryInterface.removeIndex(
+    'scalar_tap_receipts',
+    'scalar_tap_receipts_timestamp_ns_idx',
+  )
+  await queryInterface.dropTable('scalar_tap_receipts')
```
gusinacio, line 50: This should be just a "removeColumn"

packages/indexer-agent/src/db/migrations/16-modify-invalid-receipts-table.ts
```diff
@@ -0,0 +1,51 @@
...
+  logger.debug(
+    `Modifying tables scalar_tap_receipts_invalid to add extra column`,
+  )
+
+  if (tables.includes('scalar_tap_receipts_invalid')) {
+    await queryInterface.addColumn('scalar_tap_receipts_invalid', 'error_log', {
+      type: DataTypes.TEXT,
+      allowNull: false,
+    })
+  }
```
gusinacio, line 26: Could you add a "default value" for the current rows? Else, this migration will break in case an indexer contains any invalid receipt.

Also, could you add a default value for new inserts so this is a non-breaking change?

gusinacio (changes requested):

gusinacio (approved):

dwerner (approved):
looks good to me

gusinacio:
This PR unblocks https://github.com/graphprotocol/indexer-rs/issues/272

PR #991, opened 2024-08-26
indexer-service: remove LRU max length

dwerner:
Removes the limit on the LRU cache. This should prevent request times from spiking and we'll see any issues in instance RAM utilization instead.


fordN (approved):
👍

PR #995, opened 2024-08-28
fix: use subgraph timestamp for tap

### LINKED ISSUE #993: Pending RAVs doesn't take into account sender_address
Pending RAVs filters only uses `allocationID`s to filter and decide how to update on the database, but ravs are unique given a tuple (sender, allocation) so we need to update it to not have problems with multiple senders.

### LINKED ISSUE #994: RAVs retry is breaking if subgraph is late
The current code expect that subgraph is up to date at all moments. This is breaking on the following scenario:

1. RAV is redeemed and marked as redeemed.
2. Subgraph is late, so it doesn't return the rav transaction
3. Retry system thinks that it was a reorg and remove the "redeemed flag"
4. System retries it but it fails with "rav already redeemed"

We must fix this issue and indexers that already have a rav on this state should be able to recover itself by:
- If the RAV transaction already exists, update the rav's "redeemed flag".

gusinacio:
Fixes #993
Fixes #994

### DIFF
--- a/packages/indexer-common/src/allocations/__tests__/tap.test.ts
+++ b/packages/indexer-common/src/allocations/__tests__/tap.test.ts
@@ -4,9 +4,10 @@ import {
   GraphNode,
   Network,
   QueryFeeModels,
-  QueryResult,
+  TapSubgraphResponse,
 } from '@graphprotocol/indexer-common'
 import {
+  Address,
   connectDatabase,
   createLogger,
   createMetrics,
@@ -15,7 +16,7 @@ import {
   toAddress,
 } from '@graphprotocol/common-ts'
 import { testNetworkSpecification } from '../../indexer-management/__tests__/util'
-import { Sequelize } from 'sequelize'
+import { Op, Sequelize } from 'sequelize'
 import { utils } from 'ethers'
 
 // Make global Jest variables available
@@ -63,8 +64,17 @@ const setup = async () => {
   receiptCollector = network.receiptCollector
 }
 
+const ALLOCATION_ID_1 = toAddress('edde47df40c29949a75a6693c77834c00b8ad626')
+const ALLOCATION_ID_2 = toAddress('dead47df40c29949a75a6693c77834c00b8ad624')
+const ALLOCATION_ID_3 = toAddress('6aea8894b5ab5a36cdc2d8be9290046801dd5fed')
+
+const SENDER_ADDRESS_1 = toAddress('ffcf8fdee72ac11b5c542428b35eef5769c409f0')
+const SENDER_ADDRESS_2 = toAddress('dead47df40c29949a75a6693c77834c00b8ad624')
+const SENDER_ADDRESS_3 = toAddress('6aea8894b5ab5a36cdc2d8be9290046801dd5fed')
+
+// last rav not redeemed
 const rav = {
-  allocationId: toAddress('edde47df40c29949a75a6693c77834c00b8ad626'),
+  allocationId: ALLOCATION_ID_1,
   last: true,
   final: false,
   timestampNs: 1709067401177959664n,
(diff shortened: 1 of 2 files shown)


carlosvdr (approved):
sheesh nice, LGTM

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -690,7 +690,11 @@ export class AllocationReceiptCollector implements ReceiptCollector {
         ),
       },
     )
-    return response.data!
+    if (!response.data) {
```
carlosvdr, line 693: This was added in case an undefined or None is returned?
Cause it can return an empty response

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -125,25 +151,209 @@ describe('TAP', () => {
     timeout,
   )
 
+  test('should revert the rav request', async () => {
```
aasseman, line 154: Would that be more accurate?
```suggestion
  test('`revertRavsRedeemed` should revert RAV redeem status in DB only if older than subgraph last block', async () => {
```

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -125,25 +151,209 @@ describe('TAP', () => {
...
+        final: false,
+        redeemedAt: {
+          [Op.ne]: null,
+        },
+      },
+    })
+    expect(lastRedeemedRavs).toEqual([])
+  })
+
+  test('should not revert the rav request, allocation_id not in the list ', async () => {
```
aasseman, line 237:
```suggestion
  test('`revertRavsRedeemed` should not revert the RAV redeem status in DB if (allocation, sender) not in the revert list ', async () => {
```

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -125,25 +151,209 @@ describe('TAP', () => {
...
+    // we have a redeemed non-final rav in our database
```
aasseman, line 238: Is this comment to be removed?

packages/indexer-common/src/allocations/__tests__/tap.test.ts
```diff
@@ -245,68 +453,72 @@ describe('TAP', () => {
   )
 
   test(
-    'test mark final rav',
+    'test mark final rav via filterAndUpdateRavs',
```
aasseman, line 456:
```suggestion
    'test mark final rav via `filterAndUpdateRavs`',
```

aasseman (changes requested):

packages/indexer-common/src/allocations/query-fees.ts
gusinacio: It should only return empty response in case there was an error. In that case, we want to throw an error.

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -178,9 +193,13 @@ export class AllocationReceiptCollector implements ReceiptCollector {
     // flag during startup.
     collector.startReceiptCollecting()
     collector.startVoucherProcessing()
-    if (collector.tapContracts) {
+    if (collector.tapContracts && collector.tapSubgraph) {
       collector.logger.info(`RAV processing is initiated`)
       collector.startRAVProcessing()
+    } else {
+      collector.logger.info(`RAV process not initiated. 
+        Tap Contracts: ${!!collector.tapSubgraph}.
```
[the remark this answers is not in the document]
gusinacio: Done

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -564,73 +588,158 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
+    return await this.models.receiptAggregateVouchers.findAll({
       where: { last: true, final: false },
     })
-    // Obtain allocationIds to use as filter in subgraph
-    const unfinalizedRavsAllocationIds = unfinalizedRAVs.map((rav) =>
-      rav.getSignedRAV().rav.allocationId.toLowerCase(),
+  }
+
+  private async filterAndUpdateRavs(
+    ravLastNotFinal: ReceiptAggregateVoucher[],
```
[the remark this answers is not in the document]
gusinacio: Done

packages/indexer-common/src/allocations/query-fees.ts
```diff
@@ -564,73 +588,158 @@ export class AllocationReceiptCollector implements ReceiptCollector {
...
-      // https://github.com/sequelize/sequelize/issues/7664 (bug been open for 7 years no fix yet or ever)
+    // we use the subgraph timestamp to make decisions
+    // block timestamp minus 1 minute (because of blockchain timestamp uncertainty)
+    const ONE_MINUTE = 60
+    const blockTimestampSecs = tapSubgraphResponse._meta.block.timestamp - ONE_MINUTE
+
+    // Mark RAVs as unredeemed in DB if the TAP subgraph couldn't find the redeem Tx.
+    // To handle a chain reorg that "unredeemed" the RAVs.
+    // WE use sql directly due to a bug in sequelize update:
+    // https://github.com/sequelize/sequelize/issues/7664 (bug been open for 7 years no fix yet or ever)
```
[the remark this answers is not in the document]
gusinacio: Thanks, done

packages/indexer-common/src/allocations/__tests__/tap.test.ts
gusinacio: Seems more accurate indeed, updated!

packages/indexer-common/src/allocations/__tests__/tap.test.ts
gusinacio: Done

packages/indexer-common/src/allocations/__tests__/tap.test.ts
gusinacio: I usually create english sentenses to structure the test, and then implement the code. But this felt disconnected, gonna remove it.

packages/indexer-common/src/allocations/__tests__/tap.test.ts
gusinacio: done

aasseman (approved):
LGTM. Tested in prod and it redeems RAVs fine.

dwerner (approved):

PR #1013, opened 2024-10-04
fix: add pagination to tap collector

### LINKED ISSUE #1011: Add pagination to transaction retry logic
We have a system that checks for reverts and marks a transaction as final.

The problem happens in case someone has more than 100 redeem transactions pending. The system doesn't receive the transactions on this query because of pagination limits and assumes that it was reverted. It retries and the indexer keeps receiving a "Failed to redeem RAV" error message.

Here is the code where pagination logic must be implemented. https://github.com/graphprotocol/indexer/blob/ade405e07dd46e30dc4e7807417974133b87f2b0/packages/indexer-common/src/allocations/query-fees.ts#L656-L690

gusinacio:
Fixes #1011 

Depends on #1012

### DIFF
--- a/packages/indexer-common/src/allocations/__tests__/tap-pagination.test.ts
+++ b/packages/indexer-common/src/allocations/__tests__/tap-pagination.test.ts
@@ -72,56 +72,10 @@ function paginateArray<T>(
   return array.slice(startIndex, startIndex + pageSize)
 }
 
-const mockQueryTapSubgraph = jest
-  .fn()
-  .mockImplementation(async (_, variables): Promise<QueryResult<TapSubgraphResponse>> => {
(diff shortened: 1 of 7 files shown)


carlosvdr:
Looks good but Id like seeing some sort of test or let us know if any test was done on your side for this

dwerner:
> Looks good but Id like seeing some sort of test or let us know if any test was done on your side for this

@gusinacio was this addressed?

gusinacio:
Hello @dwerner and @carlosvdr. I just tested everything. Could you review again?

carlosvdr (approved):
Nice this tests are great

dwerner (approved):

PR #1015, opened 2024-10-07
feat: check escrow balance before rav redeem

### LINKED ISSUE #1009: Don't redeem RAV if there's no escrow balance
In case indexers remove the denylist from an specific sender, indexer-agent shouldn't burn RAVs by sending them all at once.

We need a way to make sure that we only try to redeem RAVs if we have balance on their escrow account.

gusinacio:
Fixes #1009

### DIFF
--- a/packages/indexer-common/src/allocations/__tests__/escrow-accounts.test.ts
+++ b/packages/indexer-common/src/allocations/__tests__/escrow-accounts.test.ts
@@ -1,0 +1,67 @@
+import { Address, toAddress } from '@graphprotocol/common-ts'
+import { EscrowAccountResponse, EscrowAccounts } from '../escrow-accounts'
+
+const timeout = 30000
+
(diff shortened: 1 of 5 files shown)


carlosvdr:
Looks good to me, but was this tested in any way?

carlosvdr:
If there is any way to test through testnet or a local network would be nice to make sure this doesnt crash

carlosvdr:
Perhaps some unit test 
Where you mock the tap Subgraph response for the escrow balance
To get the expected outcomes
One where it wont pass due to it not being enough and another test where it continues the desired path

gusinacio:
Hey @carlosvdr and @dwerner could you give another review here?

carlosvdr:
Oh yes this is nice! Works for me , great job :)

carlosvdr (approved):

dwerner (approved):

PR #1037, opened 2024-11-05
Use badges to redirect to a release

gusinacio:


carlosvdr (approved):
this is pretty nice , cool

fordN:
Love this automation! Thanks @gusinacio & @carlosvdr 

This change the release process and directly affects very important indexer docs referenced every day, so would great to get more buy-in from everyone before merging this type of change in the future.

gusinacio:
Hey, thanks for letting me know about this. It became a big issue with confused indexers running older versions. At the same time, we got a thumbs-up from @trader-payne in Discord so I thought it would be good to merge.

Next time I'll bring this for more people to discuss.

trader-payne:
@fordN we've been talking about an automated way of pointing to the Qm hashes and versions since two years now, and even though we had discussions about ENS or using data edge subgraphs for this, nobody took actions. This is what we've come up with in the interim. I think it's a great feature and everyone I've asked so far seems pretty happy with it. 

The only thing we have to keep in mind going forward is to make sure we don't create releases until things are production ready. It was the same case before anyway, so nothing really changed if you ask me.

fordN:
Right, I'm not commenting on the solution. I love the automation. ❤️  

It's just something we now have to be aware of in process (changes process and assumptions a bit), so we all need to be aware of it before executing the change.

fordN:
@trader-payne Which I clearly stated in my original message, so I don't get why you are framing my comment as being against automation or against the solution itself.

trader-payne:
Makes sense @fordN, my bad! I'm half responsible for this merge as I asked Gustavo to do it once I saw easy it was to solve the problem 😬 But yeah you're right!

PR #1046, opened 2024-11-18
common: add pagination support to allocation queries

dwerner:


packages/indexer-common/src/indexer-management/resolvers/allocations.ts
```diff
@@ -203,27 +223,43 @@ async function queryAllocations(
...
+      ...filterVars,
+      lastId,
+    }
+    const result = await networkSubgraph.checkedQuery(
+      ALLOCATION_QUERIES[filterType],
+      pageVars,
+    )
+
+    if (result.error) {
+      logger.warning('Query failed', {
```
fordN, line 240: Maybe we can be a bit more descriptive about what failed here?

fordN (approved):
Looks good! 👍 

As follow on work we could explore generalizing this queryAllocations() function and reusing elsewhere since we use similar code in `indexer-common/src/allocations/monitor.ts`.

PR #1056, opened 2024-12-11
common: fix subgraph pagination

dwerner:
Bump subgraph query page size to 1000 as per #1058 

Also fixes subgraph pagination that sometimes results in duplicates in this test:

https://github.com/graphprotocol/indexer/blob/main/packages/indexer-common/src/indexer-management/__tests__/helpers.test.ts#L355

Which is leading to this failure in tests sometimes:

https://github.com/graphprotocol/indexer/actions/runs/12281805281/job/34271522211?pr=1055

```
@graphprotocol/indexer-common:   ● Monitor: CI › Fetch subgraphs
@graphprotocol/indexer-common:     expect(received).toEqual(expected) // deep equality
@graphprotocol/indexer-common:     Expected: 106
@graphprotocol/indexer-common:     Received: 107
@graphprotocol/indexer-common: 
@graphprotocol/indexer-common:       461 |       'HGZgSEpRbwY3H54vyrUfqn8RXqDA23qT9yLqyxZXZTNW',
@graphprotocol/indexer-common:       462 |     ])
@graphprotocol/indexer-common:     > 463 |     await expect(subgraphs.length).toEqual(106)
@graphprotocol/indexer-common:           |                                    ^
@graphprotocol/indexer-common:       464 |   })
```

PR #1058, opened 2024-12-12
fix: optimize the round trips when reconciling the subgraph deployments 

abarmat:
## Issue

I was running the Indexer Agent and after a few days I realized the amount of queries it sent to the network subgraphs was out of proportion and quickly draining my api key.

## **Investigation**

The Indexer Agent performs a number of “housekeeping” queries at regular periods of time to sync with the latest network status.

One of those queries gets the entire set of subgraphs on the network but it does so paginating by 10 items batches. With the current number of subgraphs over 10,000, it takes more than 1,000 queries to get them all.

The eventual runs every 120 seconds, that means it's sending about 720,000 queries a day.

## Resolution

Set the `first` parameter to `1000` for the query within `subgraphDeployments` function of the `NetworkMonitor.`

### DIFF
--- a/packages/indexer-common/src/indexer-management/monitor.ts
+++ b/packages/indexer-common/src/indexer-management/monitor.ts
@@ -652,7 +652,7 @@ export class NetworkMonitor {
     const deployments: SubgraphDeployment[] = []
     const queryProgress = {
       lastId: '',
-      first: 10,
+      first: 1000,
       fetched: 0,
       exhausted: false,
       retriesRemaining: 10,


dwerner (approved):
Good find. I have another PR in that was intended to fix a test in that same function (#1056), I bumped the page size there as well.

fordN (approved):
🙌

dwerner:
closing in favor of https://github.com/graphprotocol/indexer/pull/1056, thanks!

abarmat:
Also tagging @fordN for visibility

dwerner:
My bad, I mixed up a similar fix.

PR #1059, opened 2024-12-12
common: skip tests that make rpc calls

dwerner:
For now, disabling tests that make RPC calls which are blocking up CI anytime that fails.


fordN (approved):
pourquoi pas? 👍

PR #1063, opened 2024-12-13
*: Add support for local querying of Epoch and TAP subgraphs

fordN:
## Changes
- Merge `EpochSubgraph`, `NetworkSubgraph`, and `TAPSubgraph` class implementations into a single, more generalized class: `SubgraphClient`. By doing so support for local deployment querying is added for Epoch and TAP subgraphs.
- Update startup of agent to ensure Epoch and TAP subgraphs are syncing locally (along with network subgraph).
- Update docs to include deployment id startup options for Epoch and TAP subgraphs.

## Background
The goal here is to add support for querying local deployments of the Epoch and TAP subgraphs. Instead of updating the EpochSubgraph and TAPSubgraph classes to that effect I merged those in with the NetworkSubgraph class which already has local querying support and auto switching between local and external based on health of endpoints.


dwerner (approved):
:+1:

carlosvdr:
Looks pretty good, I love the idea of having one more generalized class.
My only question is, if you were able to test this, either on a testnet or somewhere to make sure it isn't breaking anything?

gusinacio (approved):
Great addition, we really needed this feature of indexing tap subgraph.

fordN:
@carlosvdr still need to test on a running indexer! I won't merge this until I do so.

PR #1065, opened 2024-12-18
common, agent: switch most timer calls to sequential impl

dwerner:
This PR adds two new functions which implement a sequential timer with back-pressure, as well as using those in place of where we previously used a `timer(...).tryMap|pipe|map|reduce`.

A long standing issue with `timer()` is that it's blind to the work done further down the combinator pipeline. So if the work in a timer took longer than the next `setInterval` backing the timer to complete, we could potentially start executing the next iteration before completing the one we are currently on. 

These newly introduced functions watch the amount of work done in their `Mapper` or `Reducer` functions respectively.

I'm considering this a stop-gap solution as it provides a single logged function which still makes use of eventuals.


gusinacio (approved):
LGTM. Great work, this will make indexer-agent way more stable.

PR #1069, opened 2024-12-26
common: change getPendingRAVs PAGE_SIZE from 1000 to 200, fix http 413

dwerner:


dwerner:
Verified that this solves receipt collection on the upgrade indexer.

rotarur (approved):

PR #1070, opened 2024-12-26
issue: action-batch

dwerner:
This PR is an optimization for where the graph-node has thousands of active deployments, and calling the `indexerStatus` endpoint is taking a long time for the upgrade indexer.


fordN (approved):
🚀 👍

PR #1074, opened 2025-01-09
mark actions as PENDING while executing

dwerner:
This PR marks actions as PENDING while they are being executed. Out of this we can remove the dependency on a long-lived database transaction that can time out and fail while execution may have succeeded.

This attempts to address a couple of issues that have shown up lately:

1. Contract execution fails sometimes, maybe due to estimateGas failing or if an allocation is already present for an action being executed (stale network state)
2. A long-live database transaction has timed out, leading to issue 1. after the next execution of a batch.

The behavior I've opted for here is: success and failures during execution result in the same behavior as before, however if there is an _exception_ during execution, (such as from interacting with graph-node or from the contracts), then the Action is returned to APPROVED state and will be processed again on the next execution batch.

One edge case this introduces is that if the process were to be interrupted while Actions are marked PENDING, then those will be left in that state. A warning is issued to the operator that these will need to be examined and cleared manually.

### DIFF
--- a/packages/indexer-common/src/indexer-management/actions.ts
+++ b/packages/indexer-common/src/indexer-management/actions.ts
@@ -206,7 +206,41 @@ export class ActionManager {
     })
   }
 
-  private async updateActionStatuses(
+  /**
+   * Mark actions with the given status.
+   * @param actions
+   * @param transaction
+   * @param status
+   * @returns updated actions
+   * @throws error if the update fails
+   */
+  private async markActions(
+    actions: Action[],
+    transaction: Transaction,
+    status: ActionStatus,
+  ): Promise<Action[]> {
+    const ids = actions.map((action) => action.id)
+    const [, updatedActions] = await this.models.Action.update(
+      {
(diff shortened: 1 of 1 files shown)


packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -206,7 +206,44 @@ export class ActionManager {
...
+   * @returns updated actions
+   * @throws error if the update fails
+   */
+  private async markActions(
+    actions: Action[],
+    transaction: Transaction,
+    status: ActionStatus,
+  ): Promise<Action[]> {
+    const updatedActions: Action[] = []
+    for (const result of actions) {
```
fordN, line 223: What do you think about doing a bulk update SQL tx with array of ids instead of for loop that updates one at a time?

fordN (commented):
Looks good! Just one question about efficiency of updating statuses.

fordN (approved):
👍

Issue #1071, opened 2024-12-26
Allow closing allocations when maxBlock is set in manifest


dwerner:
Previously, subgraphs continued scanning even when all the data sources had passed the end block.
This update finds the maximum end block across the data sources and stops the subgraph from scanning once the max_end_block is reached.

The agent won't be able to close allocations for those subgraphs because they aren't at chainhead.

related: https://github.com/graphprotocol/graph-node/pull/5583

dwerner:
Reproducing this with local-network:

1. create a subgraph with endBlock defined (1000 in my case)
2. deploy and allocate to it
3. mine blocks past the end block defined
4. verify that latestBlock in indexerStatus is remaining at 1000 while chainHeadBlock keeps on going
5. queue up an unallocate and wait for the action to be executed
6. observe this error below:

```json
{
    "action": 3,
    "component": "AllocationManager",
    "error": {
        "cause": "POI not available for deployment at current epoch start block.\n              currentEpochStartBlock: 1210\n              deploymentStatus: [{\"subgraphDeployment\":{\"kind\":\"deployment-id\",\"value\":\"0x07e7cae9b5004024f70ad7d0f473b5b1867a7698fc871b9816e625a210d751fd\"},\"synced\":true,\"health\":\"healthy\",\"fatalError\":null,\"node\":\"default\",\"chains\":[{\"network\":\"hardhat\",\"latestBlock\":{\"number\":\"1000\",\"hash\":\"6a7425df7ae6b0e3cba250a85b1b370e49c05e1b08910732903731b8479af38c\"},\"chainHeadBlock\":{\"number\":\"1211\",\"hash\":\"d7487e29ab1280039235fc9d0faab92863cd85dc475a6660351fa20f44efa3c6\"},\"earliestBlock\":{\"number\":\"0\",\"hash\":\"0x0\"}}]}]",
        "code": "IE067",
        "explanation": "https://github.com/graphprotocol/indexer/blob/main/docs/errors.md#ie067"
    },
    "formatted_time": "2025-02-04 15:43:15:223",
    "hostname": "c829539864db",
    "level": 50,
    "level_text": "ERROR",
    "msg": "Failed to prepare tx call data",
    "name": "IndexerAgent",
    "pid": 173,
    "protocolNetwork": "eip155:1337",
    "time": 1738712595223
}
```

dwerner:
[...]

PR #1085, opened 2025-02-05
fix: closes issue #1071 - fetch a poi from the latestBlock if it's behind chainhead

dwerner:
From the arbitration charter:

> If a subgraph encounters an error that prevents it from being synced up to a the first block of the epoch, then an Indexer should submit the last valid PoI for the block before the error occured. Indexers may continue to collect indexing rewards on a broken subgraph until the market indicates it is no longer useful to index that subgraph and removes all signal from the subgraph.

This PR implements checking if the PoI exists at the last block that the subgraph is synced to when trying to unallocate, if we can't find a PoI in the 'first-block-of-current-epoch' and the subgraph synced `lastBlock` as reported by graph node is behind chainhead.

### DIFF
--- a/packages/indexer-common/src/indexer-management/monitor.ts
+++ b/packages/indexer-common/src/indexer-management/monitor.ts
@@ -18,6 +18,7 @@ import {
   resolveChainAlias,
   TransferredSubgraphDeployment,
   sequentialTimerReduce,
+  IndexingStatus,
 } from '@graphprotocol/indexer-common'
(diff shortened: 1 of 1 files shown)


abarmat:
I came across this issue with this subgraph https://graphseer.com/subgraph-deployment/Qme97Z18ey3xYQfKQjs13UwbwYiiUCUvYHrSKcXuPAAHpD/overview

PR #1090, opened 2025-02-07
common,cli,agent: remove multi-network-mode

dwerner:
This PR removes multi-network mode while preserving the network config parsing logic that it was tied to. This enables a significant reduction in complexity.

Changes:
- Now the agent expects there to only be a single yaml file in the `network-specifications-dir`
- If `network-specifications-dir` is provided then the values are used, otherwise single-network mode's style of pulling values from argv/env are used.
- Transfer (L1 -> L2) code is removed since the transfer tool is removed now.
- the env var INDEXER_AGENT_MULTINETWORK_MODE is ignored


packages/indexer-common/src/indexer-management/actions.ts
```diff
@@ -147,62 +142,52 @@ export class ActionManager {
...
-      await this.multiNetworks.mapNetworkMapped(
-        approvedActionsByNetwork,
-        async (network: Network, approvedActions: Action[]) => {
-          const networkLogger = logger.child({
-            protocolNetwork: network.specification.networkIdentifier,
-            indexer: network.specification.indexerOptions.address,
-            operator: network.transactionManager.wallet.address,
+      const networkLogger = logger.child({
+        protocolNetwork: this.network.specification.networkIdentifier,
+        indexer: this.network.specification.indexerOptions.address,
```
pcarranzav, line 148: might it still be useful to log the operator address here?

packages/indexer-common/src/indexer-management/resolvers/allocations.ts
```diff
@@ -298,67 +297,68 @@ async function queryAllocations(
...
+      latestBlock: currentEpoch.latestBlock,
+      maxAllocationEpochs,
+      blocksPerEpoch: epochLength.toNumber(),
+      avgBlockTime: 13000,
+      protocolNetwork: network.specification.networkIdentifier,
+    }
 
-        return queryAllocations(logger, networkSubgraph, variables, context)
-      },
+    const allocationsByNetwork = await queryAllocations(
```
pcarranzav, line 357: nit: same as above, `allocationsByNetwork` might be a confusing name? maybe just `allocationsResult`?

packages/indexer-agent/src/agent.ts
```diff
@@ -329,8 +260,12 @@ export class Agent {
       sequentialTimerMap(
         { logger, milliseconds: requestIntervalLarge },
         async () => {
-          if (this.deploymentManagement === DeploymentManagementMode.AUTO) {
-            logger.debug('Fetching active deployments')
+          const { network } = this.networkAndOperator
+          if (
+            this.deploymentManagement === DeploymentManagementMode.AUTO ||
+            network.networkMonitor.poiDisputeMonitoringEnabled()
```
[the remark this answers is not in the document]
dwerner: No so this was an optimization I made for the upgrade indexer which was querying all (some 25k) deployments from the graph node in this eventual, despite not needing to. Neither `DeploymentManagementMode.AUTO` or PoiDisputeMonitoring were enabled there so it made sense to block the call behind that. IIRC In a later testing session I found that when `DeploymentManagementMode.AUTO` was enabled, this still wasn't being hit due to the eventual in multiNetwork and networkMapped not having a value at the right time. So this is re-enabling the optimization to the narrower case, but now without the networkMapped hoop to jump through.

pcarranzav (commented):
Finished reviewing, just a couple more details

PR #1099, opened 2025-02-21
common: also add chain aliases stripped of their -mainnet suffix

dwerner:
This supports shorter network aliases stripped of their -mainnet suffix, if it's present.


fordN (approved):
looks good!

matthewdarwin:
You could add aliases in the registry instead of having differing list here.

dwerner:
> You could add aliases in the registry instead of having differing list here.

Agreed that would be ideal. I'll loop back around in the future and put in a PR there.

PR #1104, opened 2025-02-28
common: use protocolNetwork configured

dwerner:


fordN (commented):
nice, what's here looks good! 👍  
We'll also want to remove protocolNetwork requirement from actions commands such as https://github.com/graphprotocol/indexer/blob/main/packages/indexer-cli/src/commands/indexer/actions/queue.ts#L76.

fordN (approved):

PR #1106, opened 2025-03-06
common, cli: resolves #1049 - remove complex cost model support

### LINKED ISSUE #1049: Remove complex cost model support
As of release [v24.2.0](https://github.com/edgeandnode/gateway/releases/tag/v24.2.0), the gateway no longer supports "complex" cost models. The gateway only supports cost models of the form `default => x;`, where `x` is a literal value. To avoid user confusion and reduce maintenance costs, I suggest the following changes:
- [x] Remove the DAI injection feature. Based on a review of cost models from active indexers, this feature is not in use.
- [x] Remove the `indexer cost set/get variables` commands from the CLI
- [x] Replace the `indexer cost set/get model` with `indexer cost set/get`, where the user can only set a float value. At this point the `@graphprotocol/cost-model` dependency can be removed.

dwerner:


Theodus (approved):
LGTM. This is great!

PR #1114, opened 2025-04-17
common: rule processing for performance

madumas:
This change significantly improves the performance of rule matching operations by implementing a Map-based lookup system. Previously, each deployment required a linear search through all rules, resulting in O(n*m) complexity for n deployments and m rules.

Key changes:
- Implemented Map-based rule lookup for O(1) access to deployment rules
- Added caching of global rule to avoid repeated searches
- Introduced `PreprocessedRules` interface to maintain type safety
- Optimized both `evaluateDeployments` and `matchingRuleExists` to use the new lookup system

This optimization is particularly impactful for large rule sets, reducing the complexity from O(n*m) to O(n) where n is the number of deployments. For example, with 800 rules and 15,000 deployments, the number of operations is reduced from approximately 12 million to 15,000.


madumas:
CI jobs appear to be failing due to permissions or because I'm outside the org

dwerner (approved):

PR #1118, opened 2025-05-20
chore: add is horizon ready flag to network monitor

Maikol:


packages/indexer-common/src/transactions.ts
```diff
@@ -420,6 +423,42 @@ export class TransactionManager {
     })
   }
 
+  async monitorIsHorizon(
```
tmigone, line 426: I think this is not necessary. It seems to me that the monitor functions on `TransactionManager` are not being used at all.

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -1086,6 +1086,43 @@ Please submit an issue at https://github.com/graphprotocol/block-oracle/issues/n
...
+    logger: Logger,
+    contracts: GraphHorizonContracts & SubgraphServiceContracts,
+    interval: number = 300_000,
+  ): Promise<Eventual<boolean>> {
+    // Get initial value
+
+    const initialValue = await contracts.HorizonStaking.getMaxThawingPeriod()
+      .then((maxThawingPeriod) => maxThawingPeriod > 0)
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      .catch((_) => false)
```
tmigone, line 1099: This is so that it's compatible with pre-horizon no?

packages/indexer-common/src/transactions.ts
tmigone: I can't see why we have the other monitor functions here, perhaps it's dead code that can be removed? I would not add the horizon monitor here.

tmigone (changes requested):
looks good, just that comment on transactions.ts

packages/indexer-common/src/transactions.ts
tmigone: seems to me the `TransactionsManager` handles write ops and the `NetworkMonitor` handles read ops, it's likely previously we had just one for both and the code got duplicated?

packages/indexer-common/src/indexer-management/monitor.ts
Maikol: Yup

tmigone (approved):

PR #1126, opened 2025-05-26
Horizon: provision management

tmigone:
Signed-off-by: Tomás Migone <tomas@edgeandnode.com>

### DIFF
--- a/packages/indexer-common/src/indexer-management/resolvers/provisions.ts
+++ b/packages/indexer-common/src/indexer-management/resolvers/provisions.ts
@@ -1,0 +1,128 @@
+import { Network } from '@graphprotocol/indexer-common'
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/ban-types */
+
+import gql from 'graphql-tag'
+
+import {
+  IndexerManagementResolverContext,
+} from '@graphprotocol/indexer-common'
+import { extractNetwork } from './utils'
+
+enum ProvisionQuery {
+  all = 'all',
+}
+
+interface ProvisionInfo {
+  id: string
+  dataService: string
+  indexer: string
+  tokensProvisioned: string
+  tokensAllocated: string
+  tokensThawing: string
+  maxVerifierCut: string
+  thawingPeriod: string
+  protocolNetwork: string
+}
+
+const PROVISION_QUERIES = {
+  // No need to paginate, there can only be one provision per (indexer, dataService) pair
+  [ProvisionQuery.all]: gql`
+    query provisions($indexer: String!, $dataService: String!) {
+      provisions(
+        where: { 
+          indexer: $indexer,
+          dataService: $dataService
+        }
+        orderBy: id
+        orderDirection: asc
+        first: 1000
+      ) {
+        id
+        indexer {
+          id
+        }
+        dataService {
+          id
+        }
+        tokensProvisioned
+        tokensAllocated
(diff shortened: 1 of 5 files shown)


packages/indexer-common/src/indexer-management/resolvers/provisions.ts
```diff
@@ -0,0 +1,644 @@
...
+      logger.info(`Successfully added stake to provision`, {
+        amountGRT: formatGRT(addToProvisionEventLogs.tokens),
+        transaction: receipt.hash,
+      })
+
+      return {
+        id: provision.id,
+        dataService,
+        indexer,
+        tokensProvisioned: (provision.tokensProvisioned + provisionAmount).toString(), // TODO: we could re-fetch the provision instead
```
Maikol, line 290: Not a big but to me the title `tokensProvisioned` isn't clear if it's the new provision size or the tokens added.

packages/indexer-common/src/indexer-management/resolvers/provisions.ts
```diff
@@ -0,0 +1,644 @@
...
+
+    if (!(await network.isHorizon.value())) {
+      throw indexerError(IndexerErrorCode.IE082)
+    }
+
+    const indexer = network.specification.indexerOptions.address.toLowerCase()
+    const dataService = contracts.SubgraphService.target.toString().toLowerCase()
+    const thawAmount = parseGRT(amount)
+
+    if (thawAmount < 0n) {
```
Maikol, line 335: Why not check that it's `> 0n`? we can't thaw 0 tokens on Staking.

packages/indexer-common/src/indexer-management/resolvers/provisions.ts
```diff
@@ -0,0 +1,644 @@
...
+    try {
+      // Check if the provision exists - this will throw if it doesn't
+      const provision = await networkMonitor.provision(indexer, dataService)
+
+      logger.debug('Provision found', {
+        provision,
+      })
+
+      const delegationRatio = await contracts.SubgraphService.getDelegationRatio()
+      const tokensAvailable = await contracts.HorizonStaking.getTokensAvailable(
```
Maikol, line 354: Should this be `getProviderTokensAvailable`? I believe we don't need the delegation ratio.

packages/indexer-common/src/indexer-management/resolvers/provisions.ts
```diff
@@ -0,0 +1,644 @@
...
+    const networkMonitor = network.networkMonitor
+    const contracts = network.contracts
+    const transactionManager = network.transactionManager
+
+    if (!(await network.isHorizon.value())) {
+      throw indexerError(IndexerErrorCode.IE082)
+    }
+
+    const indexer = network.specification.indexerOptions.address.toLowerCase()
+    const dataService = contracts.SubgraphService.target.toString().toLowerCase()
```
Maikol, line 332: Why do we lowercase? Should I be using lowercased?

tmigone: It only matters if it's a subgraph query as those are case sensitive. For on chain interactions via RPC calls it doesn't matter.

packages/indexer-common/src/indexer-management/resolvers/provisions.ts
tmigone: ahhyeah good catch. you can't thaw delegator's tokens :D

PR #1131, opened 2025-06-24
Release Branch for v0.24.x

dwerner:


tap-contracts.json
tmigone, line 1: i think this is a leftover from the horizon branch?

tmigone (approved):

tap-contracts.json
dwerner: ty

PR #1133, opened 2025-07-16
batch RAV redeem

dwerner:


dwerner:
this is fixed in horizon

PR #1136, opened 2025-08-14
fix: ensure allocation keys derive from root wallet

Maikol:


tmigone (approved):

tmigone:
Amazing catch! looks like ethers v5 uses `"m"` as default derivation while v6 uses `"m/44'/60'/0'/0/0"`.

PR #1154, opened 2025-12-16
fix: handle over allocated condition correctly

tmigone:
### Problem

When an indexer is over-allocated (allocated tokens exceed provisioned stake + delegated stake), the unallocate operation fails the simulation with `AllocationClosed` (0x61b66e0d) error. This happens because:

1. The indexer constructs a multicall: [`collect()`, `stopService()`] (`stopService()` is the new `closeAllocation()`)
2. During `collect()`, the contract presents POI, collects rewards and then checks for over allocated condition
3. If over-allocated, it automatically calls `_closeAllocation()` closing the allocation
4. Then `stopService()` executes and tries to close an already-closed allocation → reverts

```
"Failed to execute subgraph service transaction: execution reverted (unknown custom error) (action=\"estimateGas\", data=\"0x61b66e0d000000000000000000000000cf8e12d86245fdc7937ac38331cd41bcf06a1264000000000000000000000000000000000000000000000000000000006940768f\"
```

### Solution

Before building the unallocate transaction, check `SubgraphService.isOverAllocated(allocationId)`:
  - If over-allocated: Only call `collect()` — it will auto-close the allocation
  - If not over-allocated: Use normal multicall [`collect()`, `stopService()`]

Note that it's possible for the indexer to be over allocated when the agent checks for the condition, but under allocated once the transaction (consisting of just `collect()`) is confirmed on chain. In this case, on chain the indexer would have collected rewards but the allocation would remain open. The agent handles this well, it will expect an `AllocationClosed` event regardless of the calls being made, if it does not find one it will mark the action as failed. Retrying it later will result in the allocation being closed correctly (and potentially new rewards being collected as well).

Also fixed event detection for `graph indexer allocation close` command to look for `AllocationClosed` instead of `ServiceStopped`, since `AllocationClosed` is emitted in both paths.

### DIFF
--- a/docs/networks/arbitrum-sepolia.md
+++ b/docs/networks/arbitrum-sepolia.md
@@ -7,8 +7,8 @@ The Graph Network's testnet is on Arbitrum Sepolia (eip155:421614). Sepolia netw
 | Component          | Release                                                                              |
 | ------------------ | ------------------------------------------------------------------------------------ |
 | contracts          | [5.3.3](https://github.com/graphprotocol/contracts/releases/tag/v5.3.3)              |
-| indexer-agent      | [0.25.0](https://github.com/graphprotocol/indexer/releases/tag/v0.25.0)              |
-| indexer-cli        | [0.25.0](https://github.com/graphprotocol/indexer/releases/tag/v0.25.0)              |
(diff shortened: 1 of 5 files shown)


dwerner (approved):
lgtm

PR #1155, opened 2025-12-17
docs: add more docs on modes and rules

tmigone:


dwerner (approved):
lgtm! Thanks! Note there is some discussion on management modes in docs/action-queue.md, but this is centered around the action queue.

PR #1172, opened 2026-02-25
[DIPS] Cherry-pick old commits + contracts setup

Maikol:


packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -0,0 +1,527 @@
...
+      this.logger.error(`Error cancelling agreement ${agreement.id}`, { error })
+    }
+  }
+  // Update the current and last allocation ids for an agreement if it exists
+  async tryUpdateAgreementAllocation(
+    deploymentId: string,
+    oldAllocationId: Address | null,
+    newAllocationId: Address | null,
+  ) {
+    const agreement = await this.models.IndexingAgreement.findOne({
```
MoonBoi9001, line 119: If two active agreements exist for the same deployment (e.g. different payers), only one gets its allocation tracking updated. Consider using findAll here, or adding cancelled_at: null to the where clause.

packages/indexer-agent/src/db/migrations/19-add-dips-to-decision-basis.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  logger.info('Migration completed')
+}
+
+export async function down({ context }: Context): Promise<void> {
+  const { queryInterface, logger } = context
+
+  logger.info('Removing dips from decision basis')
+  await queryInterface.sequelize.query(
+    `ALTER TYPE "enum_IndexingRules_decisionBasis" DROP VALUE 'dips'`,
```
MoonBoi9001, line 34: PostgreSQL doesn't support ALTER TYPE ... DROP VALUE.

https://www.postgresql.org/docs/current/datatype-enum.html#DATATYPE-ENUM-IMPLEMENTATION-DETAILS

MoonBoi9001 (approved):

packages/indexer-common/src/indexing-fees/dips.ts
Maikol: Good point. I think there might be more things to adjust for supporting multiple agreements. For MVP I think we could enforce a single agreement per subgraph. But we should still try to get it as close to supporting multiple so I can fix this in a later PR.

packages/indexer-agent/src/db/migrations/19-add-dips-to-decision-basis.ts
Maikol: Nice, we should probably clean this migration before merging. I'll add a task to do this.

PR #1173, opened 2026-03-03
[DIPS] add `pending_rca_proposals` migration

MoonBoi9001:
## Motivation

The indexer-rs DIPs flow stores validated RCA proposals in a `pending_rca_proposals` table for the agent to later pick up and accept on-chain. By convention in this codebase, the indexer-service does not run database migrations -- it defers to the agent to avoid DDL conflicts from two processes sharing a database. Without this migration, indexer-service rejects every valid proposal with `relation "pending_rca_proposals" does not exist`.

## Summary

- Adds Umzug migration `23-add-pending-rca-proposals.ts` that creates the `pending_rca_proposals` table with the schema defined in `indexer-rs/migrations/20260302000000_dips_pending_proposals.up.sql`
- The table is intentionally minimal (6 columns): `id`, `signed_payload`, `version`, `status`, `created_at`, `updated_at`. The signed payload blob is the source of truth -- the agent decodes it when accepting on-chain
- Does **not** drop the old `indexing_agreements` / `dips_receipts` tables (the existing agent DIPs code still references them; updating that code is a separate piece of work)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/packages/indexer-agent/src/db/migrations/23-add-pending-rca-proposals.ts
+++ b/packages/indexer-agent/src/db/migrations/23-add-pending-rca-proposals.ts
@@ -16,7 +16,9 @@ export async function up({ context }: Context): Promise<void> {
   const tables = await queryInterface.showAllTables()
 
   if (tables.includes('pending_rca_proposals')) {
-    logger.debug('pending_rca_proposals already exists, migration not necessary')
+    logger.debug(
+      'pending_rca_proposals already exists, migration not necessary',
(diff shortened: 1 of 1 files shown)


Maikol (commented):
I have this migration as part of https://github.com/graphprotocol/indexer/pull/1174 Don't mind if you want to merge this but we should create the index using sequilize api

PR #1174, opened 2026-03-03
feat: add pending RCA proposal consumer and migration

Maikol:
Add PendingRcaConsumer to read pending RCA proposals from the shared pending_rca_proposals table (populated by indexer-rs), decode SignedRCA bytes via toolshed, and feed decoded proposals into the agent's reconciliation loop for indexing rule creation.

- Sequelize model (read-only, defined after sync to avoid schema ownership)
- PendingRcaConsumer class with getPendingProposals/markAccepted/markRejected
- DecodedRcaProposal interface with all decoded fields
- ensureAgreementRules branches between new RCA path and legacy fallback
- Migration 23 creates pending_rca_proposals table with IF NOT EXISTS guard
- Threading: start.ts -> client.ts -> actions.ts -> allocations.ts -> dips.ts
- Unit tests with real ABI encode/decode round-trips

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -716,6 +716,11 @@ export class Agent {
               if (!operator.dipsManager) {
                 throw new Error('DipsManager is not available')
               }
+
+              await operator.dipsManager.acceptPendingProposals(
+                activeAllocations,
+              )
+
               this.logger.debug(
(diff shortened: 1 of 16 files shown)


MoonBoi9001 (approved):
`packages/indexer-agent/src/db/migrations/23-add-pending-rca-proposals.ts` Should be renamed as it collides with `packages/indexer-agent/src/db/migrations/23-actions-add-new-types.ts`

PR #1175, opened 2026-03-04
feat: implement on-chain accept for recurring agreements

Maikol:


packages/indexer-common/src/network-specification.ts
```diff
@@ -74,7 +74,7 @@ export const IndexerOptions = z
     legacyMnemonics: z.array(z.string()).default([]),
     enableDips: z.boolean().default(false),
     dipperEndpoint: z.string().url().optional(),
-    dipsAllocationAmount: GRT().default(1),
+    dipsAllocationAmount: GRT().default(0),
```
MoonBoi9001, line 77: Nit:
The CLI option `--dips-allocation-amount` still has default: 1.
Could sync the CLI default to 0 to match

MoonBoi9001 (approved):

packages/indexer-common/src/network-specification.ts
Maikol: Yes, I'll update on a follow up PR.

PR #1178, opened 2026-03-12
feat: on-chain acceptance fixes for DIPs recurring agreements

MoonBoi9001:
## Motivation

The indexer-agent's DIPs on-chain acceptance path had four issues preventing agreements from being accepted on-chain. These were discovered during local end-to-end testing of the full DIPs pipeline (dipper -> IISA -> proposal -> indexer-service -> agent -> SubgraphService contract).

## Summary

**1. Wrong contract in getRewards call**

`AllocationManager.stakeUsageSummary()` called `RewardsManager.getRewards(HorizonStaking, allocationId)` but `HorizonStaking` is not a registered rewards issuer -- `SubgraphService` is. Every allocation operation (including DIPs acceptance) reverted with "Not a rewards issuer".

**2. Acceptance latency exceeded RCA deadline**

Proposal acceptance was tied to the agent's 120s reconciliation loop, requiring 2+ cycles (240s+) before attempting on-chain acceptance. The RCA deadline is 300s, leaving insufficient slack. Added a dedicated `startProposalAcceptanceLoop()` that polls `pending_rca_proposals` every 5 seconds, independent of the reconciliation cycle.

**3. Opaque contract revert errors**

When on-chain acceptance failed, the log showed `error: null` because `tryParseCustomError` couldn't decode unknown custom errors. Added full revert context (reason, data, message, contract target) to the rejection log so the root cause is immediately visible.

**4. Indexing rule race condition**

The 5s acceptance loop added in item 2 above runs on a separate cadence from the agent's ~15s reconciliation tick, which created a race. `processProposal` clears the `pending_rca_proposals` row as soon as the acceptance receipt lands, but the `dips` indexing rule was only upserted by `ensureAgreementRules` on the next reconcile tick by iterating that same table. On fast chains the receipt consistently beats the next reconcile tick, so graph-node never gets told to deploy the subgraph and the agent spins on unallocate attempts for the just-created DIPs allocation.

Extracted the per-proposal rule logic into `ensureDipsRuleForProposal` and call it from `processProposal` before `executeTransaction(acceptIndexingAgreement)` fires. Reconciliation still calls the same helper as defense-in-depth.

Test metadata version updated to `0` to match `IndexingAgreementVersion.V1` (first Solidity enum variant = 0). See companion PRs: edgeandnode/dipper#583, graphprotocol/indexer-rs#983.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -717,9 +717,8 @@ export class Agent {
                 throw new Error('DipsManager is not available')
               }
 
-              await operator.dipsManager.acceptPendingProposals(
-                activeAllocations,
-              )
+              // Proposal acceptance is handled by the dedicated fast loop
+              // (startProposalAcceptanceLoop), not the reconciliation cycle.
 
               this.logger.debug(
                 `Matching agreement allocations for network ${network.specification.networkIdentifier}`,
--- a/packages/indexer-common/src/indexer-management/allocations.ts
+++ b/packages/indexer-common/src/indexer-management/allocations.ts
@@ -178,6 +178,7 @@ export class AllocationManager {
         this,
         this.pendingRcaModel,
       )
+      this.dipsManager.startProposalAcceptanceLoop()
     }
   }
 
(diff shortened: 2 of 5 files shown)


Maikol (approved):

MoonBoi9001:
Closing to reopen with a clearer framing around the dual-loop architecture. The scope drifted too far from what was originally reviewed and the commit history became noisy. Reopening on a fresh branch with the fast-path loop as the headline feature.

PR #1181, opened 2026-03-17
chore: remove reallocate step and add continuous RAV collection for long lived allocation

Maikol:


packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -57,6 +57,27 @@ export class NetworkMonitor {
     private epochSubgraph: SubgraphClient,
   ) {}
 
+  async hasActiveDipsAgreement(allocationId: string): Promise<boolean> {
```
MoonBoi9001, line 60: I ran this query against network sub and got `Type 'Query' has no field 'indexingAgreements'` just making a note here, although likely doesnt matter anyway as we have the other subgraph. The note here is more for incase this doesnt get looked at again later.

packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -57,6 +57,27 @@ export class NetworkMonitor {
...
+              first: 1
+            ) {
+              id
+            }
+          }
+        `,
+        { allocationId: allocationId.toLowerCase() },
+      )
+      return (result.data?.indexingAgreements?.length ?? 0) > 0
+    } catch {
```
MoonBoi9001, line 76: Suggestion: After fixing the above query, we could change `catch { return false }` to `catch { return true }` and log a warning so the failure is observable.

At the moment if the error path is hit because the query doesnt work, then we can't determine agreement status, but the unallocate action goes through anyway.

packages/indexer-common/src/allocations/graph-tally-collector.ts
```diff
@@ -414,11 +450,10 @@ export class GraphTallyCollector {
     )
   }
 
-  // redeem only if last is true
-  // Later can add order and limit
+  // Fetch all non-finalized RAVs (both active and closed allocations)
   private async pendingRAVs(): Promise<ReceiptAggregateVoucherV2[]> {
     return await this.models.receiptAggregateVouchersV2.findAll({
-      where: { last: true, final: false },
+      where: { final: false },
```
MoonBoi9001, line 456: Dropping `last: true` also pulls in old RAVs that dont need to be collected since their value is already included in the latest RAV.

Suggested: `{ last: true, final: false }`

`markRavsAsFinal` requires last = TRUE, so old RAVs can't be archived.

packages/indexer-common/src/allocations/graph-tally-collector.ts
```diff
@@ -472,7 +507,7 @@ export class GraphTallyCollector {
     await this.markRavsAsFinal(blockTimestampSecs)
 
     return await this.models.receiptAggregateVouchersV2.findAll({
-      where: { redeemedAt: null, final: false, last: true },
+      where: { final: false },
```
MoonBoi9001, line 510: same as above

packages/indexer-common/src/allocations/graph-tally-collector.ts
```diff
@@ -934,19 +969,33 @@ export class GraphTallyCollector {
...
 
-          try {
-            await this.markRavAsRedeemed(rav.rav.collectionId, rav.rav.payer)
-            logger.debug('[TAPv2] RAV marked as redeemed', {
+          const allocationId = collectionIdToAllocationId(rav.rav.collectionId)
+          const isActive = await this.isActiveAllocation(allocationId)
+
+          if (isActive) {
+            // Active allocation: update cooldown, don't mark as redeemed
+            this.lastCollectedAt.set(rav.rav.collectionId.toLowerCase(), Date.now())
```
MoonBoi9001, line 977: Nit: `lastCollectedAt` entries are added for active allocations, but never removed when the allocation closes.

MoonBoi9001 (approved):

packages/indexer-common/src/allocations/graph-tally-collector.ts
Maikol: RAVs are re-generated for the same allocation, we don't get duplicated RAVs but it overrides the existing one. If we filter by `last: true` we won't be redeeming until the indexer closes the allocation and allocations are now long lived so we need to not filter by `last: true`.

packages/indexer-common/src/indexer-management/monitor.ts
Maikol: yes, this is changing later, there's a branch for the network subgraph that has this entity but this is already different in #1201

packages/indexer-common/src/indexer-management/monitor.ts
Maikol: this is a critical error for running dips, if dips is enabled I believe we should stop execution until fixed, else you could be running without realizing that you have a wrong subgraph configured unless you look at the logs and you could easily miss it due to the amount of output the agent generates.

PR #1185, opened 2026-03-23
feat: implement on-chain collect for DIPs agreements

Maikol:


packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -470,6 +481,146 @@ export class DipsManager {
...
+
+    logger.info(
+      `${readyIds.length} of ${agreements.length} agreement(s) ready for collection`,
+    )
+
+    const readyAgreements = agreements.filter((a) => readyIds.includes(a.id))
+
+    for (const agreement of readyAgreements) {
+      try {
+        await this.tryCollectAgreement(agreement, blockNumber, logger)
```
MoonBoi9001: `tryCollectAgreement` returns early on `'paused'`/`'unauthorized'` without throwing, so `updateAfterCollection` runs as if the collect succeeded, when when no tx was actually sent.

packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -470,6 +481,146 @@ export class DipsManager {
...
+        this.collectionTracker.updateAfterCollection(agreement.id, nowSeconds)
+      } catch (err) {
+        if (this.isDeterministicError(err)) {
+          const parsedError = tryParseCustomError(err)
+          logger.warn('Deterministic error collecting agreement, skipping', {
```
MoonBoi9001: Note: A deterministic revert here logs `warn` and skips, so the agreement is retried forever.

packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -470,6 +481,146 @@ export class DipsManager {
...
+  private async tryCollectAgreement(
+    agreement: SubgraphIndexingAgreement,
+    blockNumber: number,
+    logger: Logger,
+  ): Promise<void> {
+    const deploymentId = new SubgraphDeploymentID(agreement.subgraphDeploymentId)
+    const entityCounts = await this.graphNode.entityCount([deploymentId])
+    const entities = entityCounts[0]
+
+    const recentBlock = blockNumber - 10
```
MoonBoi9001: Could extract into RECENT_BLOCK_OFFSET or similar

MoonBoi9001 (approved):

PR #1190, opened 2026-04-03
feat: implement on-chain cancel for DIPs agreements

Maikol:


packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -481,6 +481,99 @@ export class DipsManager {
...
+
+    for (const agreement of agreements) {
+      const subgraphDeploymentID = new SubgraphDeploymentID(
+        agreement.subgraphDeploymentId,
+      )
+      const blocklistedRule = allDeploymentRules.find(
+        (rule) =>
+          new SubgraphDeploymentID(rule.identifier).bytes32 ===
+            subgraphDeploymentID.bytes32 &&
+          rule.decisionBasis === IndexingDecisionBasis.NEVER,
```
MoonBoi9001: should we also match OFFCHAIN?

packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -547,6 +644,26 @@ export class DipsManager {
...
+    const isCancelledByPayer = agreement.state === 3
+    const isExpired =
+      Number(agreement.endsAt) > 0 && Number(agreement.endsAt) < nowSeconds
+
+    if (isCancelledByPayer || isExpired) {
+      logger.info('Agreement finished, removing from collection tracker', {
+        agreementId: agreement.id,
+        reason: isCancelledByPayer ? 'payer-cancelled' : 'expired',
+      })
+      this.collectionTracker.remove(agreement.id)
```
MoonBoi9001, line 668: Do we need to collect fees between indexers `lastCollectionAt` and payer's cancellation event before we remove the agreement from the tracker?

MoonBoi9001 (approved):

packages/indexer-common/src/indexing-fees/dips.ts
Maikol: this flow runs after the agreement is canceled, so it's safe to remove here.

PR #1200, opened 2026-04-22
feat: speed up DIPs on-chain acceptance under backlog

MoonBoi9001:
## TL;DR

Under a backlog, DIPs agreements were timing out before the agent accepted them on-chain, because acceptance only ran on the agent's slow two-minute housekeeping cycle. This gives acceptance its own fast loop that works through pending agreements in parallel, so they land well inside the window they stay open for. It also keeps a just-accepted agreement's indexing rule in place until the network's record of that agreement catches up.

## Motivation

The indexer-agent accepts DIPs agreements on-chain, but today it only does that on its general housekeeping cycle, which runs every two minutes. On a fresh deploy with a backlog, working through the pending agreements takes several of those cycles — longer than the five-minute window an agreement stays open for — so agreements expire before they are accepted and the indexer loses the work.

Accepting in parallel also means the agent writes down the indexing rule for an agreement the moment it is accepted, before the network's published record of that agreement is visible to the agent. The same housekeeping cycle removes rules that have no matching agreement, so without care it would delete that brand-new rule and undo the acceptance; this change keeps a just-accepted rule in place until the network's record catches up.

## Summary

- Acceptance moves to its own 5-second loop instead of the agent's two-minute cycle.
- Pending agreements are accepted in parallel (up to four at once) to clear backlogs in time.
- The indexing rule is created up front so the subgraph deploys before the next cycle runs.
- A just-accepted rule is kept until the network's record of its agreement catches up.
- A specific contract-call error is treated as permanent, so the agent stops retrying it.
- Before accepting, the agent checks the offer exists on-chain (reusing the base branch's check).
- The database migration can safely run again if the value it adds is already there.

### DIFF
--- a/packages/indexer-common/src/indexing-fees/dips.ts
+++ b/packages/indexer-common/src/indexing-fees/dips.ts
@@ -8,7 +8,6 @@ import {
   Allocation,
   AllocationManager,
   AllocationStatus,
-  DipsReceiptStatus,
   GraphNode,
   IndexerManagementModels,
   IndexingDecisionBasis,
--- a/packages/indexer-common/src/indexing-fees/dips.ts
+++ b/packages/indexer-common/src/indexing-fees/dips.ts
@@ -17,6 +17,7 @@ import {
   SubgraphIdentifierType,
   upsertIndexingRule,
 } from '@graphprotocol/indexer-common'
+import pMap from 'p-map'
 
 import { PendingRcaProposal } from '../indexer-management/models/pending-rca-proposal'
 import { OfferMonitor } from './offer-monitor'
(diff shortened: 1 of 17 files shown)


MoonBoi9001:
Rebased on #1216

packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -28,15 +30,41 @@ import {
   SubgraphIndexingAgreement,
 } from './agreement-monitor'
 import { CollectionTracker } from './collection-tracker'
+import { sequentialTimerMap } from '../sequential-timer'
 import { OfferVerifier } from './offer-verifier'
 
+const DIPS_ACCEPTANCE_INTERVAL = 5_000
```
Maikol, line 36: should we make this values configurable?

packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -507,6 +656,13 @@ export class DipsManager {
       }
 
       await consumer.markAccepted(proposal.id)
+      // markAccepted clears the pending row, so the deployment drops out of the
+      // reconcile target set until the subgraph indexes the agreement. Shield
+      // its rule from the reaper meanwhile.
+      this.recentlyAcceptedDeployments.set(
```
Maikol, line 662: given the explanation needed for this line I would extract this to a new function and avoid the duplication in `acceptWithExistingAllocation`

Maikol (approved):

packages/indexer-common/src/indexing-fees/dips.ts
MoonBoi9001: https://github.com/graphprotocol/indexer/pull/1226

PR #1202, opened 2026-04-22
perf: create index for tap receipts

tmigone:
Add composite indexes on `tap_horizon_receipts` to improve query performance for collection based lookups. Creates `tap_horizon_receipts_collection_id` on `(collection_id, id)` and `tap_horizon_receipts_collection_text_id` on `(CAST(collection_id AS TEXT), id)`.

### DIFF
--- a/packages/indexer-agent/src/db/migrations/24-add-tap-horizon-indexes.ts
+++ b/packages/indexer-agent/src/db/migrations/24-add-tap-horizon-indexes.ts
@@ -11,6 +11,7 @@ interface Context {
 }
 
 const INDEX_NAME = 'tap_horizon_receipts_collection_id'
+const INDEX_NAME_TEXT = 'tap_horizon_receipts_collection_text_id'
 
 export async function up({ context }: Context): Promise<void> {
   const { queryInterface, logger } = context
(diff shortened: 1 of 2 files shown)


Maikol (approved):

packages/indexer-agent/src/db/migrations/24-add-tap-horizon-indexes.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+  )
+  logger.info(`Created index ${INDEX_NAME}`)
+
+  logger.info(
+    `Creating composite index ${INDEX_NAME_TEXT} on tap_horizon_receipts`,
+  )
+  await queryInterface.sequelize.query(`
+    CREATE INDEX CONCURRENTLY ${INDEX_NAME_TEXT}
+    ON tap_horizon_receipts (CAST(collection_id AS TEXT), id)
```
lutter, line 37: Once https://github.com/graphprotocol/indexer-rs/pull/1022 is merged, this index won't be needed

tmigone: indexers might not update in tandem (agent, service) so I added a note saying this can be removed once we see sufficient coverage with the indexer-rs version that makes this not required.

lutter (approved):

PR #1204, opened 2026-04-24
fix(allocations): pass SubgraphService as rewards issuer

RembrandtK:
The RewardsManager auth check requires the rewards issuer argument to equal subgraphService. Passing HorizonStaking instead triggers "Not a rewards issuer" revert during stakeUsageSummary, which blocks unallocate/reallocate/resize actions from ever being submitted.

Main branch fix: https://github.com/graphprotocol/indexer/pull/1195


RembrandtK:
I was under impression already on some branches at least, but seems not commited to `main-dips`.

MoonBoi9001 (approved):

PR #1216, opened 2026-05-14
fix: indexer collects final payment after payer-side cancel

MoonBoi9001:
## TL;DR

Stops the indexer-agent from losing payment for honest work when the payer (via dipper) cancels an agreement on-chain before the agent has its own chance to cancel. Adds the missing final collect on the already-canceled path.

## Motivation

When the payer cancels an indexing agreement on-chain (Studio shrinking the indexer set is the common case), the agent observes the resulting CanceledByPayer state, sweeps the deployment's DIPs rule, closes the allocation, and writes a NEVER rule so the deployment stays synced without being allocated to. The next collection tick is where the indexer loses money — the agent's "cancel blocklisted agreements" pass sees the NEVER rule and routes the already-canceled agreement back through the cancel function.

The cancel function tries a second on-chain cancel, the contract reverts because the agreement is already canceled, and the function returns early before reaching its own final-collection step. The indexer is never paid for the period the agreement was active, even though those funds are sitting in the payer's escrow and would have been collectable.

## Summary

- Skip already-canceled agreements when sweeping for opt-out rules.
- In the cancel function, skip the doomed on-chain step when already canceled.
- Proceed straight to the final collect so the indexer gets paid.
- Two new unit tests covering both branches.
- Tests added but not run locally — CI will run them.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/indexer-common/src/indexing-fees/dips.ts
+++ b/packages/indexer-common/src/indexing-fees/dips.ts
@@ -474,8 +474,7 @@ export class DipsManager {
       agreementId,
     })
 
-    // Step 1: Cancel on-chain (skipped if payer already canceled — a second
-    // cancel reverts on `InvalidAgreementState` and would skip the final collect).
+    // Step 1: Cancel on-chain (skipped if payer already canceled).
     const indexerAddress = this.network.specification.indexerOptions.address
     if (agreement.state === 'CanceledByPayer') {
       logger.info(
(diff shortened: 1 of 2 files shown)


packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -472,36 +474,43 @@ export class DipsManager {
       agreementId,
     })
 
-    // Step 1: Cancel on-chain
+    // Step 1: Cancel on-chain (skipped if payer already canceled — a second
+    // cancel reverts on `InvalidAgreementState` and would skip the final collect).
```
Maikol: nit: I would avoid adding:
```
— a second cancel reverts on `InvalidAgreementState` and would skip the final collect
```

packages/indexer-common/src/indexing-fees/dips.ts
```diff
@@ -536,6 +546,12 @@ export class DipsManager {
     })
 
     for (const agreement of agreements) {
+      // Already-canceled agreements need a final collect, not another cancel —
+      // the regular collection loop handles them. cancelAgreement also guards
+      // this state internally as defense-in-depth.
+      if (agreement.state === 'CanceledByPayer') {
```
Maikol: nit: I think we could expand this and only check agreements which state is active
```typescript
if (agreement.state !== 'Active') continue
```

Maikol (approved):

Maikol (approved):

PR #1225, opened 2026-05-29
feat: stop unpaid indexing after a DIPs agreement is canceled

MoonBoi9001:
## TL;DR

When a payer cancels a DIPs agreement, the indexer used to keep its allocation open indefinitely, indexing a subgraph it would no longer be paid for. This lets that allocation close once the agreement's final payment has been collected, so the indexer stops doing unpaid work and frees the stake. It waits for the final payment first, because the payment can only be collected while the allocation is still open.

## Motivation

A DIPs agreement pays an indexer to index a particular subgraph, and while the agreement is live the indexer's allocation for it is protected from being closed. When the payer cancels the agreement, that protection used to stay in place forever, so the allocation never closed and the indexer carried on indexing — with its stake locked in place — for a subgraph nobody was paying for.

The allocation can't simply close on cancellation, though: the final payment for work already done can only be collected while the allocation is still open, so closing too early would forfeit it. This change keeps the allocation protected only until that final payment has been collected, then releases it so the indexer's normal housekeeping can close it and put the stake to use elsewhere.

## Summary

- A canceled agreement's allocation stays protected only until its final payment is collected.
- Protection consults the payment contract, which reports fees as collectable until fully drained.
- Once nothing is left to collect, the allocation is freed and normal housekeeping closes it.
- An active, uncanceled agreement still protects its allocation exactly as before.
- If the collectable amount can't be read, the allocation stays protected, to avoid losing fees.

### DIFF
--- a/packages/indexer-common/src/indexer-management/monitor.ts
+++ b/packages/indexer-common/src/indexer-management/monitor.ts
@@ -62,20 +62,58 @@ export class NetworkMonitor {
     if (!this.indexingPaymentsSubgraph) {
       return false
     }
+    // An allocation is protected from closing while it still owes a DIPS payment.
+    // Closing it makes SubgraphService.collect revert (collect requires the
+    // allocation to be open), so a canceled agreement must keep its allocation
+    // alive until its final fees have been collected. Two cases protect:
+    //   - an Accepted agreement: still live, closing would cancel it on-chain;
+    //   - a payer-canceled agreement whose fees aren't fully collected yet.
     const result = await this.indexingPaymentsSubgraph.checkedQuery(
       gql`
         query indexingAgreements($allocationId: Bytes!) {
           indexingAgreements(
             where: { allocationId: $allocationId, state_in: [Accepted, CanceledByPayer] }
-            first: 1
           ) {
             id
+            state
           }
         }
       `,
       { allocationId: allocationId.toLowerCase() },
     )
-    return (result.data?.indexingAgreements?.length ?? 0) > 0
+    const agreements: { id: string; state: string }[] =
+      result.data?.indexingAgreements ?? []
+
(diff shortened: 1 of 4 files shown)


packages/indexer-common/src/indexer-management/monitor.ts
```diff
@@ -62,20 +62,58 @@ export class NetworkMonitor {
     if (!this.indexingPaymentsSubgraph) {
       return false
     }
+    // An allocation is protected from closing while it still owes a DIPS payment.
+    // Closing it makes SubgraphService.collect revert (collect requires the
+    // allocation to be open), so a canceled agreement must keep its allocation
+    // alive until its final fees have been collected. Two cases protect:
+    //   - an Accepted agreement: still live, closing would cancel it on-chain;
+    //   - a payer-canceled agreement whose fees aren't fully collected yet.
```
Maikol: nit: this entire comment feels out of place, maybe we can move to where `hasActiveDipsAgreement` is called.

packages/indexer-common/src/indexer-management/monitor.ts
Maikol: Given the changes to the implementation of this function I would rename this since it's no longer looking for active agreements but checking if the agreement is collectable.

packages/indexer-common/src/indexer-management/monitor.ts
MoonBoi9001: Addressed in follow up commits, also tightened this block

packages/indexer-common/src/indexer-management/monitor.ts
MoonBoi9001: renamed in follow up commit

Maikol (approved):

PR #1229, opened 2026-06-03
fix: collect indexing rewards once per epoch, not every 15 min

MoonBoi9001:
## TL;DR

An indexer that keeps an allocation open collects its indexing rewards periodically without closing it, and today the agent fires one of those collections roughly every fifteen minutes. Since rewards accrue continuously and a single collection per epoch still gathers everything earned in that epoch, this change spaces the collections out to at most once per epoch. It is an efficiency fix that cuts redundant gas — the total rewards an indexer earns are unchanged.

## Motivation

When an indexer keeps an allocation open for a long time, the agent periodically collects the indexing rewards it has earned without closing the allocation. Those rewards accrue continuously, so collecting once an epoch rather than every few minutes still gathers everything earned in between — the indexer just receives it in fewer, larger batches.

Today the agent re-collects on every reconciliation pass; the only thing holding it back is an unrelated fifteen-minute window that stops the same action being queued twice in quick succession, so the cadence tracks wall-clock time rather than epochs. On a network with day-long epochs that can mean up to ninety-six collections a day where one per epoch would gather the same rewards, and since every collection is its own on-chain transaction the indexer pays that gas over and over for no extra earnings.

## Summary

- Tag each rewards collection with the epoch it was made in.
- Skip any allocation already collected in the current epoch before collecting again.
- A new epoch makes the allocation eligible again, so it collects at most once per epoch.
- Leave the existing fifteen-minute throttle in place for all other action types.

### DIFF
--- a/packages/indexer-agent/src/__tests__/agent.ts
+++ b/packages/indexer-agent/src/__tests__/agent.ts
@@ -442,6 +442,7 @@ describe('reconcileDeploymentAllocationAction', () => {
       expect.anything(),
       [activeAllocations[0]],
       network,
+      10,
     )
   })
 
--- a/packages/indexer-agent/src/agent.ts
+++ b/packages/indexer-agent/src/agent.ts
@@ -8,7 +8,9 @@ import {
   timer,
 } from '@graphprotocol/common-ts'
 import {
+  ActionManager,
   ActionStatus,
+  ActionType,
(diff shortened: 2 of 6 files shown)


MoonBoi9001:
Closed to make way for https://github.com/graphprotocol/indexer/pull/1231

PR #1234, opened 2026-06-10
feat: start indexing sooner after accepting a payment agreement

MoonBoi9001:
## TL;DR

When an indexer accepts a payment agreement, the agent now starts indexing the subgraph straight away instead of waiting for the payer's on-chain offer to appear, and it holds the on-chain step for a few seconds so that offer has time to land. A new setting (default five seconds) controls that wait.

## Motivation

Today, when an indexer accepts a payment agreement off-chain, the agent only writes a pending record, and the rule that tells the indexing engine to start work is tied to the payer's on-chain offer being visible. Because that offer transaction takes a moment to land and be indexed, the agent repeatedly tries the on-chain acceptance and finds the offer not yet there, and indexing doesn't begin until it clears. That wastes attempts against a tight agreement deadline and delays the indexer doing the work it is being paid for. This change starts the indexing as soon as the off-chain acceptance is seen and waits a short, bounded few seconds before the first on-chain attempt, so far fewer attempts are wasted while the subgraph is already syncing.

## Summary

- Create the indexing rule as soon as an accepted agreement is seen, so syncing starts immediately.
- Wait a configurable few seconds before the first on-chain step, timed from the accept.
- Cap that wait below the agreement deadline so it can never make the agent miss the deadline.
- Add a setting for the wait (default five seconds) and thread it through the configuration.
- Add tests for early sync start, the wait window, deadline ordering, and restart safety.

### DIFF
--- a/packages/indexer-agent/src/commands/start.ts
+++ b/packages/indexer-agent/src/commands/start.ts
@@ -388,6 +388,17 @@ export const start = {
         required: false,
         group: 'Indexing Fees ("DIPs")',
       })
+      .option('dips-on-chain-accept-delay', {
+        description:
+          'How long after the off-chain accept to wait before the first on-chain ' +
+          'acceptIndexingAgreement attempt, in seconds. Gives the payer time to land ' +
+          "the offer() tx so fewer attempts hit 'offer not yet present'. The deployment " +
+          'starts syncing immediately regardless of this delay.',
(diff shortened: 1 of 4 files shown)


MoonBoi9001:
Turning this to draft for now, as not presently needed to move forward with DIPs

PR #1248, opened 2026-07-10
fix: collect high-value RAVs before dust and expose deferrals

MoonBoi9001:
This PR makes both TAP collectors reconcile their most valuable pending Receipt Aggregate Vouchers (RAVs) first, raises the per-pass batch from 100 to 1,000 rows, and makes deferred fees visible to operators.

Each pass, the collectors take a capped batch of pending RAVs and reconcile it against the chain: recording which were collected, finalizing settled ones, and clearing the rest for submission. The batch was 100 rows in unspecified database order, while RAVs below the redemption threshold (1 GRT by default) legitimately never leave the pending set, so permanently deferred rows could fill the batch and starve reconciliation for everything behind them. That matters more alongside #1247, which stops the agent from wrongly finalizing partially collected RAVs, so their below-threshold remainders now stay pending instead of being written off.

The parts:

- pendingRAVs orders by value descending, so dust sinks to the bottom of the batch, and warns when the batch comes back full.
- The batch limit rises from 100 to 1,000. The allocation id filter sent to the subgraph is chunked into groups of 100 per request, keeping request bodies the size they are today (larger bodies have drawn http 413 rejections from this endpoint before).
- New Prometheus gauges report the count of below-threshold RAVs deferred each pass and their total GRT; for Horizon (TAP v2) that value is what is still collectible, the RAV's total minus what the payer already collected. The existing info log gains the same remainder figure.

6 Postgres-backed tests cover the ordering, the full-batch case, and the chunked, block-pinned subgraph queries.

### DIFF
--- a/packages/indexer-common/src/allocations/graph-tally-collector.ts
+++ b/packages/indexer-common/src/allocations/graph-tally-collector.ts
@@ -37,6 +37,15 @@ const RAV_CHECK_INTERVAL_MS = 900_000
 // 1000 here was leading to http 413 request entity too large
 const PAGE_SIZE = 200
 
+// How many pending RAVs are reconciled per pass. Rows beyond this stay pending and are
+// picked up on later passes once higher value rows settle and leave the set.
+const PENDING_RAV_BATCH_SIZE = 1_000
+
+// How many allocation ids go into a single subgraph request. The id filter travels in
+// every request body, so it is chunked to keep requests the size they had when the
+// batch itself was capped at 100 rows (see the http 413 note above).
+const FILTER_CHUNK_SIZE = 100
+
 interface RavMetrics {
   ravRedeemsSuccess: Counter<string>
   ravRedeemsInvalid: Counter<string>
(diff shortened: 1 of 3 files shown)


packages/indexer-common/src/allocations/graph-tally-collector.ts
```diff
@@ -415,12 +444,21 @@ export class GraphTallyCollector {
...
   // redeem only if last is true
-  // Later can add order and limit
+  // Highest value first, so that RAVs worth collecting are never crowded out of the
+  // batch by dust that sits below the redemption threshold indefinitely.
   private async pendingRAVs(): Promise<ReceiptAggregateVoucherV2[]> {
-    return await this.models.receiptAggregateVouchersV2.findAll({
+    const ravs = await this.models.receiptAggregateVouchersV2.findAll({
       where: { last: true, final: false },
-      limit: 100,
+      order: [['valueAggregate', 'DESC']],
```
tmigone, line 452: Note that this does not necessarily order by highest collection amount given that the collection is `valueAggregate - alreadyCollected`.

tmigone: Though its probably only practical to do this and not consider what's already collected

tmigone (approved):

